PR #5, opened 2017-08-06
updated readme and tutorial data to use alpha diversity artifact

### LINKED ISSUE #1: update `paired-differences` example in readme to use a .qza as input
You mention that this is possible, but it'd be better to just use that in your example since that's the preferred way to do this (since it retains provenance where exporting the alpha diversity data wouldn't). 

```
--m-metadata-file ecam_map_maturity.txt
```

You could also link to the metadata tutorial, which has a good description of this.

nbokulich:
fixes #1


gregcaporaso (approved):

gregcaporaso (approved):

gregcaporaso:
@nbokulich, this looks good but note that there are merge conflicts that need to be resolved before this can be merged.

nbokulich:
thanks @gregcaporaso ! I fixed the conflict — should be ready to go.

PR #20, opened 2017-08-08
Issue 11

### LINKED ISSUE #11: strange results if state 1 value equals state 2 value
I think all differences should be zero, but that's not what we're seeing: 

screenshot 2017-08-07 13 25 05

nbokulich:
fixes #11 

This does everything described in #11 with the exception of calculating mean and median values when replicates are found. This is a bit too complicated to implement at this stage for two reasons: (1) at this stage we are just grabbing the IDs to compare, and the pairwise difference/distance values are computed later; (2) this would behave very differently for each method — e.g., average alpha diversity is easy but how do we calculate average distance? If you think we should still support averaging replicates in a later release, we can leave this issue open or create a dedicated issue.


q2_intervention/plugin_setup.py
```diff
@@ -55,9 +56,9 @@
                          'across which samples are paired.'),
         'individual_id_column': (
             'Metadata column containing subject IDs  to use for pairing '
-            'samples. WARNING: if duplicates exist for an individual ID at '
+            'samples. WARNING: if replicates exist for an individual ID at '
             'either state_1 or state_2, that subject will be dropped and '
-            'reported in standard output by default. Set duplicates="ignore" '
+            'reported in standard output by default. Set replicates="ignore" '
```
gregcaporaso: should this be `random` instead of `ignore`?

q2_intervention/_intervention.py
```diff
@@ -13,16 +13,19 @@
...
+                         _regplot_subplots_from_dataframe, _load_metadata,
+                         _check_inputs)
 
 
 def paired_differences(output_dir: str, metadata: qiime2.Metadata,
                        group_column: str, metric: str, state_column: str,
                        state_1: str, state_2: str, individual_id_column: str,
                        parametric: bool=False, palette: str='Set1',
-                       drop_duplicates: bool=True, table: pd.DataFrame=None
+                       drop_replicates: str='error', table: pd.DataFrame=None
```
gregcaporaso: I don't think this should be called `drop_replicates` - maybe `replicate_handling`?

q2_intervention/_utilities.py
```diff
@@ -27,28 +27,40 @@
 TEMPLATES = pkg_resources.resource_filename('q2_intervention', 'assets')
 
 
+def _check_inputs(state_1, state_2):
```
gregcaporaso: Could you rename this, maybe `_validate_state_values`, to be a little more self-documenting?

gregcaporaso (changes requested):
@nbokulich, generally the tests would go in at the same time as the new functionality. Are these in another PR right now?

q2_intervention/plugin_setup.py
nbokulich: 👍

q2_intervention/_intervention.py
nbokulich: 👍

q2_intervention/_utilities.py
nbokulich: I intend to validate more inputs in a future PR so will use `_validate_input_values` instead of `_validate_state_values`

nbokulich:
thanks @gregcaporaso ! I added a unit test and revised per your suggestions.

q2_longitudinal/_utilities.py
```diff
@@ -27,28 +27,40 @@
...
         for state_value in state_values:
             state_value = df[state_column].dtype.type(state_value)
-            individual_id = \
-                df[individual_id_column].dtype.type(individual_id)
+            individual_id = df[individual_id_column].dtype.type(individual_id)
             _state = df[state_column] == state_value
             _ind = df[individual_id_column] == individual_id
             individual_at_state_idx = group_md[_state & _ind].index
             if len(individual_at_state_idx) > 1:
                 print("Multiple values for {0} {1} at {2} {3} ({4})".format(
```
gregcaporaso, line 52: These `print` statements aren't going to be very useful here, since with some interfaces users will see them, and with others they won't. In general, `print` statements shouldn't be in any QIIME actions. Instead, these types of warnings should be included in the visualization - for example, see the yellow warning box in [this viz](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fdocs.qiime2.org%2F2017.7%2Fdata%2Ftutorials%2Fmoving-pictures%2Fcore-metrics-results%2Ffaith-pd-group-significance.qzv).

Since I know you don't have a lot of time to work on this before you leave, could you create an issue to replace these `print` statements with warnings in the visualization?

q2_longitudinal/_utilities.py
```diff
@@ -27,28 +27,40 @@
...
+                        'Replicate values for individual {0} at state {1}. '
+                        'Remove replicate values from input files or set '
+                        'replicate_handling parameter to select how '
+                        'replicates are handled.'))
+                elif replicate_handling == 'drop':
                     break
-                else:
+                elif replicate_handling == 'random':
                     individual_at_state_idx = [choice(individual_at_state_idx)]
             elif len(individual_at_state_idx) == 0:
```
gregcaporaso: Instead of ending this with `break`, could you replace `break` with `pass`, and then put `result.append(individual_at_state_idx[0])` in an `else` block?

q2_longitudinal/plugin_setup.py
```diff
@@ -72,10 +73,13 @@
...
-        'drop_duplicates': (
-            'If True, will discard all subject IDs with duplicate samples '
-            'at either state_1 or state_2. If False, will instead '
-            'choose one representative at random from among duplicates.')
+        'replicate_handling': (
+            'Choose how replicate samples are handled. If replicates are '
+            'detected, "error" causes method to fail; "drop" will discard all '
+            'subject IDs with replicate samples at either state_1 or state_2; '
+            '"random" chooses one representative at random from among '
+            'replicates; "mean" and "median" compute average values across '
```
gregcaporaso: `mean` and `median` aren't implemented, right? Can you remove those from this description?

gregcaporaso (changes requested):
Looks good @nbokulich, a couple of minor things...

nbokulich:
thanks @gregcaporaso ! The latter two changes are made as suggested.

The first change (about warning messages) has been turned into ticket #23

gregcaporaso (approved):

PR #27, opened 2017-08-14
removed between subject comparisons until we implement a proper statistical test for this comparison

nbokulich:
This is no longer an option for the `pairwise_distances` visualization. Will re-introduce in a later release, but we need to work out an appropriate statistical test for this. Not enough time before the 2017.8 release.

(tests like mann whitney and t-tests are inappropriate because they assume independence, but between-sample distances have multiple distances for each sample.)


gregcaporaso (approved):

gregcaporaso:
@nbokulich, can you resolve the merge conflict on this one?

PR #29, opened 2017-08-23
Bring into core distribution

ebolyen:
travis might actually pass on this one...


thermokarst (approved):
Looks good! Thanks @ebolyen!

PR #33, opened 2017-09-06
replace print statements with warnings

### LINKED ISSUE #23: replace print statements with warnings
see @gregcaporaso 's comment [here](https://github.com/qiime2/q2-longitudinal/pull/20#discussion-diff-132718585R52)

nbokulich:
fixes #23 

Uses collapsible alert box, so that users can collapse very long warning messages and get on to the results. 

Thanks @thermokarst for discussing this behavior and directing me to the collapse widget!

### DIFF
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -73,6 +73,7 @@ def _get_group_pairs(df, group_value, individual_id_column='SubjectID',
                      group_column='Group', state_column='time_point',
                      state_values=['1', '2'], replicate_handling='error'):
     results = []
+    errors = []
     group_members = df[group_column] == group_value
     group_md = df[group_members]
     for individual_id in set(group_md[individual_id_column]):
@@ -84,9 +85,11 @@ def _get_group_pairs(df, group_value, individual_id_column='SubjectID',
             _ind = df[individual_id_column] == individual_id
             individual_at_state_idx = group_md[_state & _ind].index
             if len(individual_at_state_idx) > 1:
-                print("Multiple values for {0} {1} at {2} {3} ({4})".format(
-                    individual_id_column, individual_id, state_column,
-                    state_value, ' '.join(map(str, individual_at_state_idx))))
+                errors.append(
+                    "Multiple values for {0} {1} at {2} {3} ({4})".format(
+                        individual_id_column, individual_id, state_column,
+                        state_value,
+                        ' '.join(map(str, individual_at_state_idx))))
                 if replicate_handling == 'error':
                     raise ValueError((
                         'Replicate values for individual {0} at state {1}. '
@@ -98,7 +101,7 @@ def _get_group_pairs(df, group_value, individual_id_column='SubjectID',
                 elif replicate_handling == 'drop':
                     pass
             elif len(individual_at_state_idx) == 0:
-                print("No values for {0} {1} at {2} {3}".format(
+                errors.append("No values for {0} {1} at {2} {3}".format(
                     individual_id_column, individual_id, state_column,
                     state_value))
                 pass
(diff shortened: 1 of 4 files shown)


q2_longitudinal/assets/index.html
```diff
@@ -2,6 +2,25 @@
 
 {% block content %}
 
+{% if errors %}
+<div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
+  <div class="panel panel-default">
+    <div class="panel-heading" role="tab" id="accordion-heading">
+      <h4 class="panel-title">
+        <a role="button" data-toggle="collapse" data-parent="#accordion" href="#table" aria-expanded="true" aria-controls="table">
```
ebolyen: Maybe call these `#warnings` instead?

q2_longitudinal/assets/index.html
```diff
@@ -2,6 +2,25 @@
...
+          WARNINGS (click here to collapse):
```
ebolyen: This looks a little extreme, `Warnings` is probably fine, but up to you

q2_longitudinal/assets/index.html
```diff
@@ -2,6 +2,25 @@
 
 {% block content %}
 
+{% if errors %}
+<div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
```
ebolyen: These ids (and the others) will probably collide with anything else that uses this in the future. Maybe make it `warnings-accordion`?

ebolyen: (id attributes, e.g. `#something` are expected to be unique on the page)

q2_longitudinal/assets/index.html
```diff
@@ -2,6 +2,25 @@
...
+  <div class="panel panel-default">
```
ebolyen: If you change `panel-default` to `panel-warning` it looks pretty good.

q2_longitudinal/assets/index.html
```diff
@@ -2,6 +2,25 @@
...
+    <div class="panel-heading" role="tab" id="accordion-heading">
+      <h4 class="panel-title">
+        <a role="button" data-toggle="collapse" data-parent="#accordion" href="#table" aria-expanded="true" aria-controls="table">
+          WARNINGS (click here to collapse):
+        </a>
+      </h4>
+    </div>
+    <div id="table" class="panel-collapse collapse in" role="tabpanel" aria-labelledby="accordion-heading">
+      <div class="alert alert-warning col-md-12">
+        <strong>{{ errors }}</strong>
```
ebolyen: It would be a little more idiomatic to do something like:

```
{% for error in errors %}
<p><strong>{{ error }}</strong></p>
{% endfor %}
```

This would also mean you don't need the join on an HTML tag in your python code which make things nice an d separate.

q2_longitudinal/_utilities.py
```diff
@@ -408,8 +411,12 @@ def _visualize(output_dir, multiple_group_test=False, pairwise_tests=False,
         plot.savefig(join(output_dir, 'plot.pdf'), bbox_inches='tight')
         plt.close('all')
 
+    if errors is not False:
```
ebolyen: `if errors:`

ebolyen (changes requested):
Some minor comments, mostly with respect to the HTML and templating

nbokulich:
thanks @ebolyen ! I have made all changes as you recommend (with the exception of the change in `utilities.py`, which is unnecessary after implementing the `for` loop in your second-to-last comment)

ebolyen (approved):

ebolyen:
Thanks @nbokulich!

PR #43, opened 2017-09-15
BUG: linear-mixed-effects failed when one variable was provided as input

### LINKED ISSUE #38: BUG: LME plots fail to generate when single variable is provided
When a single independent variable is used for LME, plots fail to generate because a single `AxesSubplot` object is generated — the current code expects multiple variables/subplots, and the ability to index these.

Key error is here:
```
File "/Users/nbokulich/miniconda3/envs/qiime2-2017.8/lib/python3.5/site-packages/q2_longitudinal/_utilities.py", line 351, in _regplot_subplots_from_dataframe
    ax=axes[num], lowess=lowess, ci=ci)
TypeError: 'AxesSubplot' object does not support indexing
```

This bug is noted in [this forum post](https://forum.qiime2.org/t/error-with-longitudinal-linear-mixed-effects/1183/5).

nbokulich:
fixes #38 

Fixed LME plotting function bug that caused failure when one variable is provided as input.

For whatever reason, `matplotlib.pyplot.subplots` returns `axes` as an array if `nrows` > 1, but as a single ax object if `nrows` = 1. I'm sure there must be a very good reason for this, but it threw me off, leading to this bug.

Also added a unit test to ensure that a single variable allows LME to run without failure.

### DIFF
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -344,15 +344,20 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
                                      group_by, lowess=False, ci=95,
                                      palette='Set1'):
     '''plot a single regplot for each group in group_by.'''
-    f, axes = plt.subplots(len(group_by), figsize=(6, 18))
+    height = 6 * len(group_by)
+    f, axes = plt.subplots(len(group_by), figsize=(6, height))
     for num in range(len(group_by)):
+        if len(group_by) > 1:
+            ax = axes[num]
+        else:
+            ax = axes
         sns.set_palette(palette)
         for group in metadata[group_by[num]].unique():
             subset = metadata[metadata[group_by[num]] == group]
             sns.regplot(state_column, metric, data=subset, fit_reg=True,
                         scatter_kws={"marker": ".", "s": 100}, label=group,
-                        ax=axes[num], lowess=lowess, ci=ci)
-        axes[num].legend(bbox_to_anchor=(1.05, 1), loc=2, borderaxespad=0.)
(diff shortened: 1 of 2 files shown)


q2_longitudinal/_utilities.py
```diff
@@ -344,15 +344,20 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
                                      group_by, lowess=False, ci=95,
                                      palette='Set1'):
     '''plot a single regplot for each group in group_by.'''
-    f, axes = plt.subplots(len(group_by), figsize=(6, 18))
+    height = 6 * len(group_by)
```
thermokarst: Minor stylistic note, we are now computing `len(group_by)` four times in the span of four lines - this would be a good candidate for splitting out into a new variable:

```python
group_by_count = len(group_by)
height = 6 * group_by_count
....
`

File: q2_longitudinal/_utilities.py [hunk 2]
```diff
@@ -344,15 +344,20 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
+    f, axes = plt.subplots(len(group_by), figsize=(6, height))
     for num in range(len(group_by)):
+        if len(group_by) > 1:
+            ax = axes[num]
+        else:
+            ax = axes
```
thermokarst: 👍

thermokarst (changes requested):
Thanks @nbokulich! One minor stylistic note (see comments). Otherwise this is GTG, thanks!

nbokulich:
Thanks @thermokarst ! Corrected as requested.

thermokarst (approved):

PR #48, opened 2017-09-20
ENH: added raw data download to visualizers

### LINKED ISSUE #47: add raw data download for all plots
see @gregcaporaso 's comment in #36 

> One more thought: Would it be worth adding a download link for the data used to generate these plots? Since we're not including any statistics, it could be useful to allow the user to get that data to do statistics on their own. If we did that, I think you'd want a tsv file that looks something like:
>```
>delivery  month  studyid  shannon
>vaginal  0  42  2.2
>cesarian  0  43  3.0
>```
>(EDITED: to make the example file tab-separated text instead of comma-separated)

nbokulich:
fixes #47 

raw data download already existed for `pairwise-differences` and `pairwise-distances`; this has been added for `linear-mixed-effects` and `volatility`.

### DIFF
--- a/q2_longitudinal/_longitudinal.py
+++ b/q2_longitudinal/_longitudinal.py
@@ -133,8 +133,12 @@ def linear_mixed_effects(output_dir: str, metadata: qiime2.Metadata,
                'Individual ID column'],
         name='Linear mixed effects parameters')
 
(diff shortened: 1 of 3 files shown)


thermokarst (approved):
LGTM! Thanks @nbokulich!

PR #55, opened 2017-09-22
ENH: added NMIT method

### LINKED ISSUE #53: ENH: new action: NMIT
https://www.ncbi.nlm.nih.gov/pubmed/28872698

nbokulich:
fixes #53 

This method generates a distance matrix based on temporal similarity between individuals sampled longitudinally. Pass to `beta-group-significance` or `emperor` to compare/view sample similarities.

Very many thanks to @elong0527 who wrote most of the core nmit utility functions.

The `ecam_map_small.txt` and `ecam-table-small.qza` files are just new tiny test files (NMIT can take a long time to run) and don't really need review.


q2_longitudinal/_utilities.py
```diff
@@ -577,3 +578,74 @@ def _stats_and_visuals(output_dir, pairs, metric, group_column,
...
+        output from temporal_corr
+    id_set: pd.Series
+        unique subject ids from individual_id with index attached
+    dist_method: str
+        method supported by scipy.linalg.norm parameter ord
+    '''
+    id_n = len(id_set)
+
+    dist = np.zeros((id_n, id_n))
+    for i, id_i in enumerate(id_set):
```
nbokulich, line 613: @elong0527 it looks like there may be two unnecessary steps in this loop:
1. this will calculate the distance of each `(id_i, id_j)` pair twice. we should instead just duplicate the result if this creates a symmetric matrix
2. this calculates the identical-sample distance along the diagonal of the matrix. Should this just be skipped so that all self-comparisons are 0?

elong0527: Yes, you are right. I will update the code to improve performance.

elong0527:
Hi Nick:

I am not sure how could I modify your updated code on github. So I put my code here. Could you help me update the code? Thanks !

Essentially, I use Pandas `groupby` function to increase efficiency. Now, `nmit` run much faster. 

That will be great if you could show me how to change code in a Pull Request.

Below is the change I propose:

```
def _temporal_corr(table, individual_id, corr_method="kendall"):
    '''Create Temporal correlation from a feature table containing repeated
    measures samples.
    table: pd.DataFrame
        rows are samples, columns are features (count / relative_abundance)
    individual_id: pd.Series
        subject id of samples, with the same length as df
    corr_method: str
        temporal correlation method, "kendall", "pearson", "spearman"
    '''

    # Start to calculate temporal correlation
    table["individual_id"] = individual_id
    results = table.groupby(["individual_id"]).corr(method=corr_method).fillna(0)

    return results


def _temporal_distance(corr, id_set, dist_method="fro"):
    '''Calculate Distance Matrix from temporal correlation data.
    corr_dict: dict
        output from temporal_corr
    id_set: pd.Series
        unique subject ids from individual_id with index attached
    dist_method: str
        method supported by scipy.linalg.norm parameter ord
    '''
    id_n = len(id_set)

    dist = np.zeros((id_n, id_n))
    for i, id_i in enumerate(id_set):
        for j, id_j in enumerate(id_set):
            if i<j:
                dist[i, j] = linalg.norm(corr.loc[id_i] - corr.loc[id_j], ord=dist_method)
                dist[j, i] = dist[i, j]
            if i==j:
                dist[i,j]  = 0
    return DistanceMatrix(dist, ids=id_set.index)
```

Below is a test code 
```
import numpy as np
import pandas as pd
from scipy import linalg
#from skbio import DistanceMatrix

t_n = 15
otu_n = 4
id_n = 2
x = np.random.rand(t_n, otu_n, id_n)
table = pd.DataFrame(x.reshape(t_n * id_n, otu_n))
individual_id = np.repeat(["s" + str(i) for i in range(id_n)], t_n)
id_set = np.unique(individual_id)
corr_method = "kendall"
results = {}

table["individual_id"] = individual_id
results = table.groupby(["individual_id"]).corr(method=corr_method).fillna(0)
corr = _temporal_corr(table, individual_id, corr_method="kendall")
_temporal_distance(corr, id_set, dist_method="fro")
```

elong0527:
I am not sure if python has a way to vectorize the distance matrix calculation step. I search online but failed because I might use some wrong keywords.

In R, there is a function called `outer`. 

https://stat.ethz.ch/R-manual/R-devel/library/base/html/outer.html

I am looking for a similar way to calculate the distance matrix in Python by dummy codes below.

```
def foo(corr, corr, dist_method):
    linalg.norm(corr.loc[id_i] - corr.loc[id_j], ord=dist_method)

outer(corr, corr, dist_method)
```

nbokulich:
@elong0527 you will need to do a PR into my branch (`nbokulich:issue-53`). Something like the following:
```
git pull https://github.com/nbokulich/q2-intervention.git issue-53
[add and commit your changes]
git push origin issue-53
```
and then issue a PR from your branch into my repo. Once the changes are merged into my branch, they will automatically appear in this PR.

I can review once I see the diff in my repo. Thanks!

Re: outer products, is [numpy.outer](https://docs.scipy.org/doc/numpy/reference/generated/numpy.outer.html) what you are looking for?

nbokulich:
@elong0527 never mind — looks like numpy.outer just does multiplication, doesn't apply a function to the outer product. However, this [SO](https://stackoverflow.com/questions/24839481/python-matrix-outer-product) led me to find that [numpy ufuncs](https://docs.scipy.org/doc/numpy-1.13.0/reference/ufuncs.html) have an [outer attribute](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ufunc.outer.html#numpy.ufunc.outer) to do what I think you are trying to do. If numpy has an appropriate ufunc, that might solve what you are trying to do.

I can investigate some more later tonight if that doesn't help... thanks!

elong0527:
We do not have to worry too much for this. The code should run relative fast for ECAM data.

q2_longitudinal/tests/data/ecam_map_small.txt
```diff
@@ -0,0 +1,19 @@
+#SampleID	antiexposedall	day_of_life	delivery	diet	diet_3	mom_child	month	month_of_life	sample_summary	sex	studyid
```
thermokarst: I don't see this file being utilized anywhere in the tests - is it necessary? Or maybe a test wasn't updated that should be using this?

tutorial_data/ecam_map_small.txt
```diff
@@ -0,0 +1,19 @@
+#SampleID	antiexposedall	day_of_life	delivery	diet	diet_3	mom_child	month	month_of_life	sample_summary	sex	studyid
```
thermokarst: Sorry I didn't notice this before, but do we need the `tutorial_data/` directory? Seems like now that these files are up on data.qiime2.org we should remove this from the repo. Thoughts?

q2_longitudinal/_longitudinal.py
```diff
@@ -172,3 +172,24 @@ def volatility(output_dir: str, metadata: qiime2.Metadata, group_column: str,
 
     _visualize(output_dir, plot=chart, summary=summary, raw_data=raw_data,
                plot_name='Control charts')
+
+
+def nmit(table: pd.DataFrame, metadata: qiime2.Metadata,
+         individual_id_column: str, corr_method="kendall", dist_method="fro"
+         ) -> DistanceMatrix:
+
+    # load and merge feature table and metadata to ensure IDs match
```
thermokarst: This sounds like we want to ensure that the biom table ids are a subset of the metadata ids, then, only process the intersection of the two (as in, the biom ids). We have an example of this kind of validation [here](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/_heatmap/_visualizer.py#L78-L82), if you are interested.

q2_longitudinal/_longitudinal.py
```diff
@@ -172,3 +172,24 @@ def volatility(output_dir: str, metadata: qiime2.Metadata, group_column: str,
...
+    metadata = _load_metadata(metadata)
+    metadata = metadata[[individual_id_column]]
+    metadata = pd.concat([metadata, table], axis=1, join='inner')
```
thermokarst: Personal preference here, I prefer to use [`pd.join`](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/_heatmap/_visualizer.py#L94). I like to use this because it borrows from database terminology, which is arguably a little more explicit than `concat`ing and various axes (a join is always a row-wise join in db-land). No need to change, just wanted to put it out there.

q2_longitudinal/_utilities.py
```diff
@@ -577,3 +578,78 @@ def _stats_and_visuals(output_dir, pairs, metric, group_column,
                paired_difference_tests, boxplot, summary=summary,
                errors=errors, plot_name=plot_name,
                pairwise_test_name=pairwise_test_name)
+
+
+def _temporal_corr(taxa, individual_id, corr_method="kendall"):
+    '''Create Temporal correlation from a feature table
+    containing repeated measures samples.
+    taxa: pd.DataFrame
```
thermokarst: It is a little strange to refer to this table as `taxa` here, maybe we could rename to `table`?

Does this method make any special assumptions about feature identifiers in the table? Is it assuming they will be taxonomic identifiers? If so, we should probably chat about this a bit.

q2_longitudinal/_utilities.py
```diff
@@ -577,3 +578,78 @@ def _stats_and_visuals(output_dir, pairs, metric, group_column,
...
+    taxa["individual_id"] = individual_id
+    results = taxa.groupby(["individual_id"]).corr(method=corr_method)
+    results = results.fillna(0)
+
+    return results
+
+
+def _temporal_distance(corr, id_set, dist_method="fro"):
+    '''Calculate Distance Matrix from temporal correlation data.
+    corr_dict: pd.DataFrame
```
thermokarst: This docstring doesn't match the signature.

q2_longitudinal/_utilities.py
thermokarst: How about something like this?

```python
dist = np.zeros((id_n, id_n))
for i, id_i in enumerate(id_set):
    for j, id_j in enumerate(id_set[:i]):
        dist[i, j] = dist[j, i] = linalg.norm(corr.loc[id_i] - corr.loc[id_j], ord=dist_method)
```

q2_longitudinal/plugin_setup.py
```diff
@@ -28,10 +28,14 @@
 )
 
 
-base_parameters = {
+basest_parameters = {
```
thermokarst: Yowzah! How about `shared_parameters`?

q2_longitudinal/plugin_setup.py
```diff
@@ -47,26 +51,32 @@
...
         **base_parameter_descriptions,
         'group_column': (
             'Metadata column on which to separate groups for comparison'),
+        'individual_id_column': (
+            'Metadata column containing subject IDs to use for pairing '
+            'samples. WARNING: if replicates exist for an individual ID at '
+            'either state_1 or state_2, that subject will be dropped and '
+            'reported in standard output by default. Set '
+            'replicate_handling="random" to instead randomly select one '
+            'member, and use --verbose to list conflicts.'),
```
thermokarst: `--verbose` is a cli-specific flag, but this method can be run in any of the QIIME 2 interfaces. Perhaps just 

> and run in verbose-mode to list conflicts.

Or something to that end.

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -171,6 +194,7 @@ def _load_dm(dm_fp):
             return dm
 
         self.table_ecam_fp = _load_features('ecam-table-maturity.qza')
+        self.table_taxa_fp = _load_features('ecam-table-small.qza')
```
thermokarst, line 197: ~see above re: `taxa` vs `table`~

thermokarst (changes requested):
Thanks @nbokulich! I have a handful of questions and comments - please let me know if you need any clarification or help! Thanks!

q2_longitudinal/tests/data/ecam_map_small.txt
nbokulich: ~~See line [197](https://github.com/nbokulich/q2-intervention/blob/issue-53/q2_longitudinal/tests/test_longitudinal.py#L197) and [282](https://github.com/nbokulich/q2-intervention/blob/issue-53/q2_longitudinal/tests/test_longitudinal.py#L282)~~

tutorial_data/ecam_map_small.txt
nbokulich: no — I suppose it is safe to remove this now. Will remove in this round of changes.

q2_longitudinal/tests/data/ecam_map_small.txt
thermokarst: Hmm, those lines aren't referencing this file, sorry if it is right in front of me...

q2_longitudinal/_longitudinal.py
nbokulich: ~~we want to make sure that the biom and metadata have identical index (sample IDs) — so filtering (of both dataframes), not validation.~~

nbokulich: 👍

q2_longitudinal/tests/data/ecam_map_small.txt
nbokulich: you are right, I have removed this file. thanks!

q2_longitudinal/_utilities.py
nbokulich: 👍

q2_longitudinal/_utilities.py
nbokulich: thanks for catching! have checked all docstrings.

q2_longitudinal/plugin_setup.py
nbokulich: 😞 I guess I was just channeling a little shakespeare or something... "shared" it is!

q2_longitudinal/plugin_setup.py
nbokulich: Thanks for catching. That's actually an old description I should have removed earlier. In actions where replicate handling occurs, a warning now appears in the viz to list samples that were dropped.

q2_longitudinal/_longitudinal.py
nbokulich: I like that too — it's shorter and more elegant than what I have, but it's also causing something to break downstream. So if this is just a matter of style, I'll leave what i have for now.

q2_longitudinal/_utilities.py
elong0527: I made the change, it is `table` originally, I am fine to use taxa/table to call it. Just to remind the user this is a OTU table :)

nbokulich:
thanks @thermokarst ! I have made all changes as suggested, with the exception of implementing `pd.join`, as mentioned above. If you decide that `pd.join` is necessary and not just stylistic, I will go back to implement/debug.

nbokulich:
@thermokarst I did get `join` working but realized that the whole `concat`/`drop` sequence was unnecessary. All we needed was some good 'ol indexing.

thermokarst:
Hi @nbokulich & @elong0527! I have made a few minor adjustments, summarized[ in this commit](https://github.com/qiime2/q2-longitudinal/pull/55/commits/d2a9568c6552ccaf80966949d89aa7928cbfb935). If this looks okay to you, please let me know.

thermokarst:
I also added [one more squash commit](https://github.com/qiime2/q2-longitudinal/pull/55/commits/d0e0d608ddd2a3687a5072934559314ef125cb34), removing the label "species" from a comment.

nbokulich:
Thanks @thermokarst ! Looks great (and thanks for the added unit test!)

thermokarst:
@elong0527, can you take a look at the changes I referenced above sometime today? We would like to have this PR merged in time for the 2017.9 release, and I think we are really close! Thanks!

elong0527:
It looks great ! Thanks @thermokarst :)

thermokarst (approved):

PR #59, opened 2017-10-04
added first_differences method

### LINKED ISSUE #58: ENH: new action: `compute-first-differences`
Computes first differences (differences in Y between sequential samples across time X)
X   Y   FD
1    1    
2   3   2
3   7   4
4   11  4

Accepts metadata files, [optionally] distance matrix or feature table

Support [value interpolation](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.interpolate.html)? Would be easy for series data but not for matrix!

nbokulich:
fixes #58 

This computes "first differences", or the difference/distance between sequential measurements taken from the same individual. This simply computes the differences, which can then be used as input for LME or other methods.


q2_longitudinal/_utilities.py
```diff
@@ -183,12 +183,15 @@ def _get_pairwise_differences(df, pairs, category, individual_id_column,
         pre_value = float(df[category][pre_idx])
         post_value = float(df[category][post_idx])
         paired_difference = post_value - pre_value
+        print(paired_difference)
```
jairideout: Can you remove this print statement?

q2_longitudinal/plugin_setup.py
```diff
@@ -28,14 +30,20 @@
...
+
 shared_parameters = {
     'metadata': Metadata,
     'individual_id_column': Str,
 }
 
 base_parameters = {
     **shared_parameters,
-    'state_column': Str,
+    **{k: v for k, v in miscellaneous_parameters.items()
```
jairideout: This approach may be clearer and avoids a loop that's only looking for a particular item:

```python
'state_column': miscellaneous_parameters['state_column']
```

If you agree with this change, can you make it here and in the other places that are looping over `miscellaneous_parameters` and `miscellaneous_parameter_descriptions`?

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +288,33 @@ def test_nmit_missing_table_ids(self):
...
+                        index=['3', '4', '5', '9', '10', '11'],
+                        name='Difference')
+        obs = first_differences(
+            metadata=qiime2.Metadata(md), state_column='Time',
+            individual_id_column='ind',
+            metric='Value', replicate_handling='drop')
+        self.assertTrue(obs.sort_index().equals(exp.sort_index()))
+
+    def test_first_differences_taxa(self):
+        first_differences(
```
jairideout: This test isn't asserting the results are correct.

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +288,33 @@ def test_nmit_missing_table_ids(self):
...
+            replicate_handling='drop', table=self.table_ecam_fp)
+
+    def test_first_differences_distance_matrix(self):
+        exp = pd.Series([0.1, 0.1, 0.3, 0.1, 0.2, 0.4],
+                        index=['3', '4', '5', '9', '10', '11'])
+        obs = first_differences(
+            metadata=qiime2.Metadata(md), state_column='Time',
+            individual_id_column='ind', replicate_handling='drop',
+            distance_matrix=dm)
+        self.assertTrue(obs.equals(exp))
```
jairideout: You can compare Series and DataFrame objects using some pandas testing utilities:

```python
import pandas.util.testing as pdt

pdt.assert_series_equal(obs, exp)
# OR:
pdt.assert_frame_equal(obs, exp)
```

This is a more common approach to asserting that pandas objects are equal, and it's handy because it'll tell you what isn't the same between the two objects. We use this strategy in several qiime2 plugins (e.g. q2-diversity), so there's some other examples floating around if it's helpful to you.

q2_longitudinal/plugin_setup.py
```diff
@@ -48,7 +56,19 @@
...
+       if k == 'replicate_handling'},
+}
+
+miscellaneous_parameter_descriptions = {
+    'state_column': ('Metadata column containing state (e.g., Time) '
+                     'across which samples are paired.'),
+    'replicate_handling': (
+        'Choose how replicate samples are handled. If replicates are '
+        'detected, "error" causes method to fail; "drop" will discard all '
+        'subject IDs with replicate samples at either state_1 or state_2; '
```
jairideout: I don't think there are `state_1` or `state_2` parameters for `first_differences`.

q2_longitudinal/plugin_setup.py
```diff
@@ -48,7 +56,19 @@
...
+        '"random" chooses one representative at random from among '
+        'replicates.')
 }
 
 shared_parameter_descriptions = {
```
jairideout: I can't comment on the exact line, but the parameter description for `metadata` (below) references `group_column`, which doesn't exist for `first_differences`.

q2_longitudinal/_longitudinal.py
```diff
@@ -192,3 +196,58 @@ def nmit(table: pd.DataFrame, metadata: qiime2.Metadata,
         corr_method=corr_method, dist_method=dist_method)
 
     return _dist
+
+
+def first_differences(metadata: qiime2.Metadata, state_column: str,
+                      individual_id_column: str, metric: str='Distance',
+                      replicate_handling: str='error',
```
jairideout: I'm getting a misformatted error message when replicates are detected (default behavior):

```

  Replicate values for individual {0} at state {1}. Remove replicate
  values from input files or set replicate_handling parameter to select
  how replicates are handled.

Debug info has been saved to /tmp/qiime2-q2cli-err-pd24umjc.log
```

The values aren't being interpolated, and perhaps it might be clearer if it read:

*Detected replicate samples for individual ...*

q2_longitudinal/plugin_setup.py
```diff
@@ -221,3 +237,39 @@
...
+plugin.methods.register_function(
+    function=first_differences,
+    inputs={'table': FeatureTable[RelativeFrequency],
+            'distance_matrix': DistanceMatrix},
+    parameters={**miscellaneous_parameters,
+                **shared_parameters,
+                'metric': Str},
+    outputs=[('first_differences', SampleData[AlphaDiversity])],
+    input_descriptions={
+        'table': 'Feature table to optionally use for paired comparisons.',
```
jairideout: Are these "paired comparisons"? I can't tell if this description is accurate or a copy-n-paste from `pairwise-differences`.

q2_longitudinal/plugin_setup.py
```diff
@@ -48,7 +56,19 @@
...
     'state_2': Str,
     'parametric': Bool,
-    'replicate_handling': Str % Choices(['error', 'random', 'drop']),
+    **{k: v for k, v in miscellaneous_parameters.items()
+       if k == 'replicate_handling'},
+}
+
+miscellaneous_parameter_descriptions = {
+    'state_column': ('Metadata column containing state (e.g., Time) '
+                     'across which samples are paired.'),
```
jairideout, line 76: Are samples being paired in `first_differences`? I can't tell if this description is accurate or a copy-n-paste from `pairwise-differences`.

q2_longitudinal/plugin_setup.py
```diff
@@ -221,3 +237,39 @@
...
+        **miscellaneous_parameter_descriptions,
+        **shared_parameter_descriptions,
+        'metric': ('Numerical metadata or artifact column to test. If '
+                   'computing first differences on a distance matrix, do not '
+                   'set a value for this parameter.'),
+    },
+    output_descriptions={'first_differences': 'Series of first differences.'},
+    name='First difference computation between sequential states',
+    description=(
+        'Calculates first differences in "metric" between sequential states '
```
jairideout, line 272: `state_column` allows both numeric and categorical metadata columns, but there isn't a guaranteed ordering to metadata columns in qiime2. How does this method order states such that they are sequential? If this column was required to be numeric, I guess the values could be sorted to form a sequence, but what happens if categorical data is supplied? I'm not sure if this is a problem or not, but wanted to bring it up in case it's a "gotcha".

q2_longitudinal/plugin_setup.py
```diff
@@ -221,3 +237,39 @@
...
+                **shared_parameters,
+                'metric': Str},
+    outputs=[('first_differences', SampleData[AlphaDiversity])],
+    input_descriptions={
+        'table': 'Feature table to optionally use for paired comparisons.',
+        'distance_matrix': 'Matrix of distances between pairs of samples.'},
+    parameter_descriptions={
+        **miscellaneous_parameter_descriptions,
+        **shared_parameter_descriptions,
+        'metric': ('Numerical metadata or artifact column to test. If '
```
jairideout: When `metric` is non-numeric can you make the error message clearer?

```

  could not convert string to float: 'n'

Debug info has been saved to /tmp/qiime2-q2cli-err-dofhig1q.log
```

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +288,33 @@ def test_nmit_missing_table_ids(self):
             nmit(table=self.table_taxa_fp, metadata=md,
                  individual_id_column='studyid')
 
+    def test_first_differences(self):
```
jairideout, line 326: General comment: these tests are pretty light, can you add some more that hit border cases and error cases? This list isn't exhaustive, but here's some examples:

- What happens when there's no states, a single state, only two states?
- What happens when there's no individuals, or a single individual?
- What happens when the `table` is empty? Single-sample only?
- What happens if the `distance_matrix` is single-sample only?
- What happens when there are no differences across states?
- Can you test different types of replicate handling other than `drop`?
- What happens if there are no replicates? 
- What happens when an ID in the feature table isn't found in the metadata? How about for a distance matrix?
- Are extra IDs in the metadata ignored when they're not found in the table or distance matrix?
- What if the same column name is supplied for `individual_id_column` and `state_column`? (should probably error).
- What happens when the same column is used for `metric` and either `individual_id_column` or `state_column`? If `table` is provided, that behavior is probably okay (but worth having a unit test for). If only `metadata` is provided, an error is probably appropriate.
- There's only one test data set being used in the unit tests. Can you add another valid test data set so that we're not overfitting the tests to a single data set?
- What happens when `metric` is non-numeric?

q2_longitudinal/_longitudinal.py
```diff
@@ -192,3 +196,58 @@ def nmit(table: pd.DataFrame, metadata: qiime2.Metadata,
         corr_method=corr_method, dist_method=dist_method)
 
     return _dist
+
+
+def first_differences(metadata: qiime2.Metadata, state_column: str,
```
jairideout, line 210: I played around with `first_differences` locally and have some suggestions for tweaking the interface. There's a few mutually-exclusive inputs to this method, so it may be worth breaking apart into separate methods that operate on metadata-only, metadata + table, and metadata + distance matrix.

If you want to keep this a single method, these changes to the interface may make it clearer (and avoid some undefined edge cases):

- Make `metric` have a default value of `None`, and error if `metric is not None` and a `distance_matrix` was provided. The current default of `Distance` is a little ambiguous because there could be a metadata column called `Distance`, or (technically) it could be a feature ID in `table`. By having a default value of `None` we're not making a statement about the metric, and can detect when the user is providing invalid input (i.e. a `distance_matrix` and a `metric`, which is disallowed).

- If `table` and `distance_matrix` are both provided, error.

- If `metric` is not provided and either `table` + `metadata` or `metadata`-only is provided, error.

- If `metric` is provided with `table` + `metadata`, the metric is looked up in `table` and errors otherwise, instead of falling back to finding the column in `metadata`.

jairideout (changes requested):
Thanks for implementing this new method @nbokulich! I didn't do a detailed review of the code yet. I focused this first-pass review on blackbox testing by pulling the PR down locally and testing it out as a user would. I have some interface changes to suggest (see comments inline). Another larger request is to add more unit tests -- I described some additional cases to test in the review comments.

Can you take a look at the review and let me know what you think? I'll hold off on doing a more in-depth review until we can finalize the interface and behavior. Thanks!

q2_longitudinal/plugin_setup.py
nbokulich: :man_facepalming: ha! thanks @jairideout ! I think I needed some more 🍵 when I wrote that...

q2_longitudinal/plugin_setup.py
nbokulich: Yes, thanks for checking — do you think this needs to be clarified further here?

q2_longitudinal/plugin_setup.py
nbokulich: thanks! You are correct — that info is out of date.

q2_longitudinal/plugin_setup.py
nbokulich: thanks! technically this method is chaining together a bunch of paired comparisons, but I have modified to make this clearer.

q2_longitudinal/plugin_setup.py
nbokulich: thanks! I've added a new validator to test this during metadata merging

q2_longitudinal/plugin_setup.py
jairideout: The word "paired" threw me off because I initially thought it meant that there were only two states (e.g. "before" and "after"). I see now that these samples are being "paired" sequentially. I don't have a suggestion to make it clearer, fine to leave as-is unless my thought process above helps prompt some text changes :)

q2_longitudinal/plugin_setup.py
nbokulich: thanks @jairideout ! This is a very good point — I think that for this method (and LME and volatility, which would be prone to the same issue) we should require state_column to be numeric... and I've added a validator (pass the data through the same validator used for the above issue). I think that's the best solution here, but am open to other ideas. For other methods, I do not want to enforce this, e.g., for paired methods.

q2_longitudinal/tests/test_longitudinal.py
nbokulich: Thanks @jairideout ! I have added a whole bunch of tests. Many of these issues (e.g., column validation) are checked by utility functions that already existed, but I have added tests for some of these anyway. I have also added several more validator functions in response to your points — you bring up a number of very good cases!

> If table is provided, that behavior is probably okay (but worth having a unit test for)

Why do you think it would be okay for metric to match other columns if the metric is in a table? I can make workarounds for that case, but for now I've added a validator to make sure *no columns* match because as far as I can think this may lead to unintended consequences.

> Can you add another valid test data set

I have one real data set and one toy data set already implemented. What kind of additional datasets would you want to see added (real or toy)?

q2_longitudinal/tests/test_longitudinal.py
nbokulich: 👍

q2_longitudinal/_longitudinal.py
nbokulich: Thanks for the feedback @jairideout — I think you are correct, it makes most sense to split this into multiple methods.

However, I would argue that splitting into 2 methods makes the most sense; one method that operates on a distance matrix and one that operates on metadata (and accepts a feature table as optional input). For the latter case, a user would always need to define `metric` whether or not a table is input, and we could always raise an error if `metric` is not a column in the table. How does that sound?

jairideout: :+1: That sounds perfect!

q2_longitudinal/plugin_setup.py
jairideout: :+1: that sounds like a good approach! With this new validator, would the other methods (LME and volatility) automatically get the fix in this PR, or are you thinking of a separate PR to update those methods? I'm fine either way, just want to know what to be on the lookout for.

nbokulich: Yep, since it's a small change I slipped that validator into LME and volatility in this PR.

q2_longitudinal/tests/test_longitudinal.py
jairideout: > Why do you think it would be okay for metric to match other columns if the metric is in a table? I can make workarounds for that case, but for now I've added a validator to make sure no columns match because as far as I can think this may lead to unintended consequences.

I was thinking it'd be technically acceptable to supply a `metric` in the feature table which could overlap with the same column name in `metadata` used for `individual_id_column` or `state_column`. We know it isn't the same column being supplied because they're coming from different files (`table` vs `metadata`). For example, imagine your `metric` in the feature table is called `foo`, and you also happen to have an `individual_id_column` called `foo`. Those `foo` names refer to columns in different files so it *could* be a valid operation, but also a very odd border case.

I think this is such a rare/odd border case, and the behavior wouldn't be clear in that situation unless the user carefully read the method's docs. The validator behavior you described is better in the long run, let's go with that!

> I have one real data set and one toy data set already implemented. What kind of additional datasets would you want to see added (real or toy)?

Apologies, I thought there was only one "valid" data set being tested, I missed the other one. I think with the addition of the data sets necessary to test those border cases I listed above, we should be GTG!

q2_longitudinal/plugin_setup.py
jairideout: Sounds good!

nbokulich:
thanks @jairideout ! I believe the newest batch of changes covers all requested changes and unit tests. 🎾 

I notice the merge conflict caused by another PR — I will fix that after you approve current changes in case other changes are required.

q2_longitudinal/_longitudinal.py
```diff
@@ -184,11 +196,51 @@ def nmit(table: pd.DataFrame, metadata: qiime2.Metadata,
...
     # run NMIT
     _dist = _nmit(
         table, metadata, individual_id_column=individual_id_column,
         corr_method=corr_method, dist_method=dist_method)
 
     return _dist
+
+
+def first_differences(metadata: qiime2.Metadata, state_column: str,
+                      individual_id_column: str, metric: str='Distance',
```
jairideout: Can you make `metric` required here?

q2_longitudinal/_longitudinal.py
```diff
@@ -184,11 +196,51 @@ def nmit(table: pd.DataFrame, metadata: qiime2.Metadata,
...
+    return _first_differences(
+        metadata, state_column, individual_id_column, metric,
+        replicate_handling, distance_matrix=None)
+
+
+def first_distances(distance_matrix: DistanceMatrix, metadata: qiime2.Metadata,
+                    state_column: str, individual_id_column: str,
+                    replicate_handling: str='error') -> pd.Series:
+
+    # find metric in metadata or derive from table and merge into metadata
```
jairideout: outdated comment

q2_longitudinal/_utilities.py
```diff
@@ -650,9 +674,71 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
 
+def _first_differences(metadata, state_column, individual_id_column, metric,
+                       replicate_handling='error', distance_matrix=None):
+
+    # let's force states to be numeric
+    _validate_is_numeric_column(metadata, state_column)
+
+    # create dummy group column in metadata so we can use downstream functions
+    # that split metadata by groups without actually bothering to do so.
+    group_column = "dummy_group"
```
jairideout: This would overwrite user data if there happens to be a column called `dummy_group` already in the metadata. Is there another way to handle this without adding a dummy column, or somehow guaranteeing a unique name that won't ever be found in the user's metadata?

q2_longitudinal/_utilities.py
```diff
@@ -650,9 +674,71 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    for s in range(len(states) - 1):
+        # get pairs of samples at each sequential state
+        group_pairs, error = _get_group_pairs(
+            metadata, group_value='null',
+            individual_id_column=individual_id_column,
+            group_column=group_column, state_column=state_column,
+            state_values=[states[s], states[s + 1]],
+            replicate_handling=replicate_handling)
+        # compute distance between pairs
+        if metric == 'Distance':
```
jairideout: I don't think this check is safe, what if the user runs `first_differences` with a `metric` that's named `'Distance'` (i.e. in their metadata file or table)? In general you'll want to avoid using sentinel values that could collide with user-defined names, otherwise the code can't tell whether the user provided the value or if it is a sentinel. This comment also relates to using `'dummy_group'` above for a sentinel.

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +310,149 @@ def test_nmit_missing_table_ids(self):
...
+            replicate_handling='drop')
+        pdt.assert_series_equal(obs, exp)
+
+    def test_first_distances_single_sample(self):
+        with self.assertRaisesRegex(RuntimeError, "Output is empty"):
+            obs = first_distances(
+                distance_matrix=dm_single_sample, metadata=qiime2.Metadata(md),
+                state_column='Time', individual_id_column='ind',
+                replicate_handling='drop')
+            print(obs)
```
jairideout: remove print statement

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +310,149 @@ def test_nmit_missing_table_ids(self):
...
+        exp.index.name = '#SampleID'
+        obs = first_distances(
+            distance_matrix=dm, metadata=qiime2.Metadata(md),
+            state_column='Time', individual_id_column='ind',
+            replicate_handling='drop')
+        pdt.assert_series_equal(obs, exp)
+
+    def test_first_distances_single_sample(self):
+        with self.assertRaisesRegex(RuntimeError, "Output is empty"):
+            obs = first_distances(
```
jairideout: This is a minor stylistic comment, up to you whether to address or leave as-is:

You don't need to store the output of a method in a variable that's expected to error, so it can be simplified to:

```python
with self.assertRaisesRegex(RuntimeError, "Output is empty"):
    first_distances(...)
```

q2_longitudinal/_utilities.py
```diff
@@ -59,16 +60,29 @@ def _validate_input_values(df, individual_id_column, group_column,
...
     for column in cols:
-        if column is not None and column not in df.columns:
+        if column not in df.columns:
             raise ValueError('{0} is not a column in your metadata'.format(
                              column))
+    # check that more than one states exist
+    if state_column is not None:
+        states = df[state_column].unique()
+        if len(states) < 2:
+            raise ValueError(("state_column must contain at least two unique "
```
jairideout: This is a minor stylistic comment, up to you whether to address or leave as-is (it applies to a few places in your PR):

You don't need to wrap string literals in parentheses if they're already contained within parentheses or any sort of braces/brackets (e.g. `[ ]`, `{ }`).

```python
raise ValueError("state_column must contain at least two unique "
                 "values. Values in {0}: {1}".format(state_column, states))
```

q2_longitudinal/_utilities.py
```diff
@@ -650,9 +674,71 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    # create dummy group column in metadata so we can use downstream functions
+    # that split metadata by groups without actually bothering to do so.
+    group_column = "dummy_group"
+    metadata[group_column] = 'null'
+
+    # calculate paired difference/distance distributions between each state
+    pairs_summary = pd.DataFrame()
+    errors = []
+    states = metadata[state_column].unique()
+    # iterate over range of sorted states in order to compare sequential states
```
jairideout: Where is the sorting happening? `pd.Series.unique()` explicitly doesn't guarantee an output order:

https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.unique.html

q2_longitudinal/plugin_setup.py
```diff
@@ -221,3 +232,68 @@
...
 )
+
+
+plugin.methods.register_function(
+    function=first_differences,
+    inputs={'table': FeatureTable[RelativeFrequency]},
+    parameters={**miscellaneous_parameters,
+                **shared_parameters,
+                'metric': Str},
+    outputs=[('first_differences', SampleData[AlphaDiversity])],
```
jairideout: Sorry I didn't notice this on the first review: these methods are producing `SampleData[AlphaDiversity]` artifacts. Conceptually, are the outputs of `first_differences` and `first_distances` alpha diversity measurements, or is this just a convenient way to store per-sample values in (i.e. because they can be viewed as metadata)?

If it's the latter case, it probably makes sense to define a new semantic type `SampleData[Something]` that captures what these values are, and that type could have `pd.Series` and `Metadata` transformers like `SampleData[AlphaDiversity]` does.

q2_longitudinal/plugin_setup.py
jairideout: I'm still getting the same error message:

```
$ qiime longitudinal first-differences --m-metadata-file ecam-sample-metadata.tsv --p-state-column month --p-individual-id-column studyid --o-first-differences out.qza --p-replicate-handling drop --p-metric antiexposedall

  could not convert string to float: 'n'

Debug info has been saved to /tmp/qiime2-q2cli-err-z3gp87xf.log
```

q2_longitudinal/plugin_setup.py
```diff
@@ -221,3 +232,68 @@
...
+    input_descriptions={
+        'table': ('Feature table to optionally use for computing first '
+                  'differences.')},
+    parameter_descriptions={
+        **miscellaneous_parameter_descriptions,
+        **shared_parameter_descriptions,
+        'metric': ('Numerical metadata or artifact column to test. If '
+                   'computing first differences on a distance matrix, do not '
```
jairideout: outdated docs now that distance matrix support is in its own method

q2_longitudinal/plugin_setup.py
```diff
@@ -221,3 +232,68 @@
...
+    parameter_descriptions={
+        **miscellaneous_parameter_descriptions,
+        **shared_parameter_descriptions,
+    },
+    output_descriptions={'first_distances': 'Series of first distances.'},
+    name='First distance computation between sequential states',
+    description=(
+        'Calculates first distances between sequential states for samples '
+        'collected from individual subjects sampled repeatedly at two or more '
+        'states. This method is similar to the first differences method, '
```
jairideout: may be clearer to quote "first differences" or refer to it as `first_differences`

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +310,149 @@ def test_nmit_missing_table_ids(self):
...
+        pdt.assert_series_equal(obs.sort_index(), exp.sort_index())
+
+    def test_first_differences_drop_duplicates(self):
+        obs = first_differences(
+            metadata=qiime2.Metadata(md_dup), state_column='Time',
+            individual_id_column='ind',
+            metric='Value', replicate_handling='random')
+        # The first diff of individual 2 is subject to random rep handling
+        mystery_number = obs.iloc[1]
+        self.assertTrue(mystery_number in [0.04999999999999999, 0.06])
```
jairideout: This check could fail on different platforms/architectures due to floating point error. Can you use [`self.assertAlmostEqual()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual) instead of exact equality?

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +310,149 @@ def test_nmit_missing_table_ids(self):
...
+
+    def test_first_differences_empty(self):
+        with self.assertRaisesRegex(ValueError, "Metadata is empty"):
+            first_differences(
+                metadata=(qiime2.Metadata(pd.DataFrame({'a': [], 'b': []}))),
+                state_column='Time', individual_id_column='ind',
+                metric='Value', replicate_handling='drop')
+
+    def test_first_differences_taxa(self):
+        first_differences(
```
jairideout: Can you assert the results are correct here?

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -288,6 +310,149 @@ def test_nmit_missing_table_ids(self):
...
+            name='Distance')
+        exp.index.name = '#SampleID'
+        obs = first_distances(
+            distance_matrix=dm,
+            metadata=qiime2.Metadata(md_one_subject_many_times),
+            state_column='Time', individual_id_column='ind',
+            replicate_handling='drop')
+        pdt.assert_series_equal(obs, exp)
+
+    # just check that it works on real data
```
jairideout: Can you assert the results somehow in this test?

q2_longitudinal/tests/test_longitudinal.py
jairideout: Thanks for adding these test cases! :+1: 

Can you also add a test case where the sequential states aren't in any kind of order in the metadata file (i.e. to test that they are sorted correctly within the method)? It'd also be good to have a test case where the order of sample IDs in the metadata file don't match the order of sample IDs in the table or distance matrix. Apologies if you already have these test cases, from a quick look I didn't see it.

jairideout (changes requested):
Thanks for the updates @nbokulich, the new tests are great! I had some more review comments after taking a second pass through.

jairideout:
Weird, my latest review got split into two reviews. Anyways, here they are!

https://github.com/qiime2/q2-longitudinal/pull/59#pullrequestreview-69979527

https://github.com/qiime2/q2-longitudinal/pull/59#pullrequestreview-70016748

q2_longitudinal/_longitudinal.py
nbokulich: 👍

q2_longitudinal/_longitudinal.py
nbokulich: 👍

q2_longitudinal/_utilities.py
nbokulich: 👍 thanks! (I fixed all cases, so you may note some utilities not related to `first-differences` appear in the next push)

q2_longitudinal/_utilities.py
nbokulich: you are right... I cannot think of an easy workaround since we need to create this dummy column in the same dataframe. It should not be an issue unless if "dummy_group" is the name of `state_column`, `individual_id_column`, or `metric`.

The best solution I can think of right now is to rename this to: `"a_name_that_no_user_of_qiime2_could_conceivably_ever_use"`

👍 👎 ❓

q2_longitudinal/_utilities.py
nbokulich: Thanks! `pd.Series.unique()` has always seemed to sort by default so thank you for revealing this. I have added sorting here and one other spot where order matters.

q2_longitudinal/_utilities.py
nbokulich: thanks! Right as always! I have changed this to the much more effective and appropriate `if distance_matrix is not None`

q2_longitudinal/plugin_setup.py
nbokulich: I have been waiting to have this conversation...

It is a bit of cases A and B. The underlying data will often be alpha diversity data and I want the outputs to be useable in many of the same ways as alpha diversity (not necessarily just as metadata). But many of the data types might not be alpha diversity data (e.g., if a taxon is input as `metric`) so I think you are right I just need to define a new semantic type.

`SampleData[FirstDifferences]`?

q2_longitudinal/tests/test_longitudinal.py
nbokulich: 👍

q2_longitudinal/tests/test_longitudinal.py
nbokulich: 😨

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/tests/test_longitudinal.py
nbokulich: 👍

q2_longitudinal/tests/test_longitudinal.py
nbokulich: 👍 I've added the expected results as a tsv

q2_longitudinal/tests/test_longitudinal.py
nbokulich: 👍 I've added the expected results as a tsv

q2_longitudinal/plugin_setup.py
nbokulich: I have defined the `SampleData[FirstDifferences]` type now

nbokulich:
thanks Jai! I have made all requested changes in both new reviews, including registering a new `SampleData[FirstDifferences]` type.

nbokulich:
Okay @jairideout I have updated the FirstDifferences format to use `__validate__` and this is ready for you to review, thanks for the guidance!

q2_longitudinal/_utilities.py
jairideout: What about calling `uuid.uuid4()` until you find a UUID that isn't a column name in the table? It's hacky but will never collide with a user's columns. We're taking this approach in `Metadata.ids()`, you could reuse that implementation here:

https://github.com/qiime2/qiime2/blob/master/qiime2/metadata.py#L294

q2_longitudinal/_longitudinal.py
```diff
@@ -184,11 +196,51 @@ def nmit(table: pd.DataFrame, metadata: qiime2.Metadata,
...
+    _validate_metadata_is_superset(
+        metadata, _tabulate_matrix_ids(distance_matrix))
+
+    # validate columns
+    # "Distance" is not actually a metadata value, so don't validate metric!
+    _validate_input_columns(
+        metadata, individual_id_column, None, state_column, None)
+
+    return _first_differences(
+        metadata, state_column, individual_id_column, metric='Distance',
```
jairideout: Can you pass `metric=None` here? `metric` is ignored when a `distance_matrix` is provided, so passing a null value might avoid confusion about `metric` actually being used to lookup values somewhere.

q2_longitudinal/plugin_setup.py
jairideout: Awesome, @ebolyen and I :+1: the name! Thanks for adding that.

q2_longitudinal/plugin_setup.py
```diff
@@ -226,3 +306,66 @@
...
+                **shared_parameters,
+                'metric': Str},
+    outputs=[('first_differences', SampleData[FirstDifferences])],
+    input_descriptions={
+        'table': ('Feature table to optionally use for computing first '
+                  'differences.')},
+    parameter_descriptions={
+        **miscellaneous_parameter_descriptions,
+        **shared_parameter_descriptions,
+        'metric': 'Numerical metadata or artifact column to test.',
```
jairideout, line 267: I think this comment got lost in the last round of review. Can you make this error message a little clearer when a non-numeric metric is supplied?

```
$ qiime longitudinal first-differences --m-metadata-file ecam-sample-metadata.tsv --p-state-column month --p-individual-id-column studyid --o-first-differences out.qza --p-replicate-handling drop --p-metric antiexposedall

  could not convert string to float: 'n'

Debug info has been saved to /tmp/qiime2-q2cli-err-z3gp87xf.log
```

q2_longitudinal/plugin_setup.py
```diff
@@ -29,14 +35,84 @@
 )
 
 
+# FirstDifferencesFormat is a near replica of AlphaDiversityFormat in q2_types
+FirstDifferences = SemanticType(
+    'FirstDifferences', variant_of=SampleData.field['type'])
+
+
+class FirstDifferencesFormat(model.TextFileFormat):
+    def _check_n_records(self, n_lines=None):
```
jairideout: If `n_lines=None` an error will be raised. Can you have it iterate through the entire file when `n_lines=None`? That'll help when the user requests `max` validation (see related comment below).

q2_longitudinal/plugin_setup.py
```diff
@@ -29,14 +35,84 @@
...
+        with self.open() as fh:
+            line = fh.readline()
```
jairideout: Can you validate the header is what's expected? i.e. `#SampleID\tDifference` or `#SampleID\tDistance`?

q2_longitudinal/plugin_setup.py
```diff
@@ -29,14 +35,84 @@
...
+            for line, _ in zip(fh, range(n_lines)):
+                cells = line.strip().split('\t')
+                if len(cells) != 2:
+                    raise ValidationError('File is not FirstDifferencesFormat')
```
jairideout: Can you describe why the file is invalid in the error message? That'll help users debug their data. The framework will wrap your error message to tell the user which format is in question, so you don't need to include that boilerplate here. Something like *"Expected two columns, found %d"* could work.

q2_longitudinal/plugin_setup.py
```diff
@@ -29,14 +35,84 @@
...
+
+    def _validate_(self, level):
+        record_count_map = {'min': 5, 'max': 1000}
```
jairideout: Can you supply `max: None` instead of `1000` to have it validate the entire file?

q2_longitudinal/plugin_setup.py
```diff
@@ -29,14 +35,84 @@
...
+    'FirstDifferences', variant_of=SampleData.field['type'])
+
+
+class FirstDifferencesFormat(model.TextFileFormat):
+    def _check_n_records(self, n_lines=None):
+        with self.open() as fh:
+            line = fh.readline()
+            for line, _ in zip(fh, range(n_lines)):
+                cells = line.strip().split('\t')
+                if len(cells) != 2:
```
jairideout: Can you also assert that the second cell of each row is a number?

q2_longitudinal/plugin_setup.py
```diff
@@ -29,14 +35,84 @@
 )
 
 
+# FirstDifferencesFormat is a near replica of AlphaDiversityFormat in q2_types
```
jairideout: Can you move all of the semantic type and format code into a new file, e.g. `_format.py` or `_type.py`? That'll help keep `plugin_setup.py` clean and matches some of the conventions we've been using elsewhere.

q2_longitudinal/_longitudinal.py
nbokulich: thanks for catching @jairideout !

q2_longitudinal/plugin_setup.py
```diff
@@ -29,14 +35,84 @@
...
+@plugin.register_transformer
+def _4(data: pd.Series) -> (FirstDifferencesFormat):
+    ff = FirstDifferencesFormat()
+    with ff.open() as fh:
+        data.to_csv(fh, sep='\t', header=True)
+    return ff
+
+
+@plugin.register_transformer
+def _5(ff: FirstDifferencesFormat) -> (pd.Series):
```
jairideout: For this transformer and the Metadata transformer below, can you cast the column to float so that the differences aren't a string dtype in the object?

q2_longitudinal/_utilities.py
nbokulich: thanks @jairideout ! I love this solution!!!

q2_longitudinal/_utilities.py
```diff
@@ -683,9 +707,71 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+
 def _validate_metadata_is_superset(metadata, table):
     metadata_ids = set(metadata.index.tolist())
     table_ids = set(table.index.tolist())
     if not table_ids.issubset(metadata_ids):
         raise ValueError('Missing samples in metadata: %r' %
                          table_ids.difference(metadata_ids))
+
+
+def _validate_is_numeric_column(metadata, metric):
```
jairideout, line 786: Can you have this utility also cast the column into a float dtype? I think it's possible for a user to supply a column that's composed of strings representing numbers -- I'm not sure what downstream code would do in that case with the "string numbers" (it might sort things lexicographically, which would be unexpected). Maybe worth a unit test too, something with states `1` and `10`?

jairideout (changes requested):
Final round of review is done, thanks for all of these changes @nbokulich! I had some more comments -- I think most are easy to address.

q2_longitudinal/_utilities.py
nbokulich: check out the new changes to `_add_metric_to_metadata`. If the metric is in metadata, it will already be cast to float by `_load_metadata`. The one metric not covered previously (I think) would have been metadata coming from feature tables.

See also new test `test_first_distances_numeric_values_represented_as_strings` to cover this.

q2_longitudinal/plugin_setup.py
nbokulich: Please @jairideout don't send me to circular import hell! 🌀🔥😱🔥🌀 maybe I could fix this in a future PR? (you could raise an issue to make sure I do it!) at this point I'm worried about getting this PR into the release...

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/plugin_setup.py
nbokulich: sorry about that! yes, I think it got buried in the split review last time.

I added a new unit test to catch this and found the spot where this was missing (I previously only raised the error when merging metric from a feature table).

nbokulich:
thanks @jairideout ! I have made all requested changes except for moving the semantic type information into a separate file. Please don't send me to circular import hell! 🌀🔥😱🔥🌀 (at least not in this PR)!

I hope all other changes look good!

jairideout:
@nbokulich I refactored the type/format/transformer code out of plugin_setup.py and made several other changes -- everything's described in [this commit](https://github.com/qiime2/q2-longitudinal/pull/59/commits/c7d5d027b0edfda9179ef7f67ed52caf2de5932c). Can you do a final review of my changes and then we can merge? Thanks!

jairideout (approved):

nbokulich:
Thank you so much for all your help on this @jairideout ! Your commit looks great to me — thank you for saving me from circular import hell! I owe you some 🍻 🌯 :1st_place_medal: 

Ready to merge when you are...

jairideout:
Thanks @nbokulich, and not a problem at all!

PR #60, opened 2017-10-04
control charts fixed label sorting and added optional xtick intervals

### LINKED ISSUE #56: ENH: volatility: user-define or automatically scale x_tick intervals
Default is to show all ticks, which gets very ugly in long, frequently sampled experiments:
image

### LINKED ISSUE #57: BUG: volatility: label sorting on plots occasionally breaks
Most plots i've seen are in correct order, but not this subplot
image
Other subplots in the same plot work, including the following that contains these same data (but different labels), so this may be a labeling issue, not a sorting issue.
image

nbokulich:
fixes #57 
fixes #56

### DIFF
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -10,6 +10,7 @@ from os.path import join
 import pandas as pd
 import seaborn as sns
 import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
 from statsmodels.sandbox.stats.multicomp import multipletests
 from itertools import combinations, cycle
 import pkg_resources
@@ -17,6 +18,7 @@ import q2templates
 from random import choice
 from statsmodels.formula.api import mixedlm
 from patsy import PatsyError
+from math import ceil
 
 import numpy as np
 from scipy import linalg
@@ -323,7 +325,8 @@ def _boxplot_from_dict(groups, hue=None, y_label=None, x_label=None,
     hue, color variables all pass directly to equivalently named
         variables in seaborn.boxplot().
     """
-    x_tick_labels = _add_sample_size_to_xtick_labels(groups)
+    x_tick_labels = [v for k, v in sorted(
+        _add_sample_size_to_xtick_labels(groups).items())]
     vals = [v for k, v in sorted(groups.items())]
 
     ax = sns.boxplot(data=vals, hue=hue, palette=palette)
@@ -336,8 +339,8 @@ def _boxplot_from_dict(groups, hue=None, y_label=None, x_label=None,
 
 def _add_sample_size_to_xtick_labels(groups):
     '''groups is a dict of lists of values.'''
-    x_tick_labels = ['{0} (n={1})'.format(k, len(v))
-                     for k, v in sorted(groups.items())]
+    x_tick_labels = {k: '{0} (n={1})'.format(k, len(v))
+                     for k, v in groups.items()}
     return x_tick_labels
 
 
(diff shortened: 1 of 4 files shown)


q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -130,7 +130,8 @@ def test_calculate_variability(self):
     def test_add_sample_size_to_xtick_labels(self):
         groups = {'a': [1, 2, 3], 'b': [1, 2], 'c': [1, 2, 3]}
         labels = _add_sample_size_to_xtick_labels(groups)
-        self.assertEqual(labels, ['a (n=3)', 'b (n=2)', 'c (n=3)'])
+        self.assertEqual(
+            sorted(labels.values()), sorted(['a (n=3)', 'b (n=2)', 'c (n=3)']))
```
thermokarst: Probably don't need to explicitly `sort` the last list (`sorted(['a (n=3)', 'b (n=2)', 'c (n=3)'])`), since it is a noop (the list order doesn't change).

q2_longitudinal/_utilities.py
```diff
@@ -392,29 +402,52 @@ def _control_chart_subplots(state_column, metric, metadata, group_column,
...
+            xtick_interval = 1
+    return xtick_interval
+
+
+def _set_xtick_labels(metadata, state_column, states, xtick_interval):
+    # pull x-axis labels from array of unique states, slice xtick_interval
+    x_tick_labels = {state: metadata[metadata[state_column] == state]
+                     for state in states[::xtick_interval]}
+    # sort labels by key (state), add sample size to labels.
+    # x_tick_labels[0] must be empty because this label does not appear in plot
```
thermokarst, line 427: Just curious, why doesn't the first label appear in the plot?

q2_longitudinal/plugin_setup.py
```diff
@@ -189,6 +191,7 @@
         'ci': 'Size of the confidence interval to plot on control chart.',
         'plot_control_limits': ('Plot global mean and control limits (2X and '
                                 '3X standard deviations).'),
+        'xtick_interval': 'Interval between major tick marks on x axis.'
```
thermokarst: Can you add a note here to indicate that some autoscaling happens? I want to make it clear to a user that things might come out in the viz slightly different than expected.

thermokarst (changes requested):
Thanks @nbokulich --- one minor addition requested (related to the param description), plus some other questions/comments. Overall looks great! Thanks!

q2_longitudinal/_utilities.py
nbokulich: To be honest I am not 100% sure... as far as I can tell, ln 438:
```
ax.xaxis.set_major_locator(ticker.MultipleLocator(xtick_interval))
```
causes the view limits to bump. Feeding text labels to a continuous x-axis causes the labels to decouple from the data, which is bad but inevitable (since we are adding sample sizes here). Just setting the ticks directly (without ``MultipleLocator``) would be ideal, but is needed here to locate the correct label position if we set `xtick_interval` > 1.

(an alternative for ln 438 is to set the xticks with something like:
```
    ax.set_xticks(np.arange(
        int(sorted(states)[0]), int(sorted(states)[-1]) + 1, xtick_interval))
```
but then we need to reset the xlimits and it just gets messy)

nbokulich:
Thanks @thermokarst ! I have made the suggested changes in the most recent commit.

thermokarst (approved):

PR #62, opened 2017-10-09
spaghetti plots enabled

### LINKED ISSUE #51: ENH: volatility: add option to plot lines for each individual's trajectory in control charts
show all individuals in control charts, colored by group membership

nbokulich:
fixes #51 

supports optional spaghetti plotting, i.e., plot vectors for each individual subject. Looks like this (faded lines are individual vectors, solid lines are what you see with spaghetti disabled):
image


q2_longitudinal/_utilities.py
```diff
@@ -400,40 +391,77 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
-    c = _set_xticks(c, metadata, state_column, states, xtick_interval)
-
     # plot individual groups' control charts
     colors = cycle(sns.color_palette(palette, n_colors=len(groups)))
-    for num in range(len(groups)):
-        group = groups[num]
-        group_md = metadata[metadata[group_column] == group]
+    cmap = {}
+    num = 1
+    for group, group_md in metadata.groupby(group_column):
```
ebolyen: `num` can turn into this:
```python
for num, (group, group_md) in enumerate(metadata.groupby(group_column), 1):
   ...
```
which will be less prone to typos

q2_longitudinal/_utilities.py
```diff
@@ -470,7 +498,7 @@ def _pointplot_from_dataframe(state_column, metric, metadata, group_by,
                               color=None, xtick_interval=None):
 
     g = sns.pointplot(data=metadata, x=state_column, y=metric, hue=group_by,
-                      ci=ci, palette=palette, color=color, ax=ax)
+                      ci=ci, palette=palette, color=color, ax=ax, markers=".")
```
ebolyen, line 509: Is this what made the dots tinier? I like it

q2_longitudinal/_utilities.py
```diff
@@ -400,40 +391,84 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
-        xtick_interval=xtick_interval)
-    c.set_title('Group volatility comparison plot')
-    c = _set_xticks(c, metadata, state_column, states, xtick_interval)
-
     # plot individual groups' control charts
     colors = cycle(sns.color_palette(palette, n_colors=len(groups)))
-    for num in range(len(groups)):
-        group = groups[num]
-        group_md = metadata[metadata[group_column] == group]
+    cmap = {}
```
ebolyen: try:
```python
cmap = dict(zip(groups, sns.color_palette(palette, n_colors=len(groups))))
```

q2_longitudinal/_utilities.py
```diff
@@ -400,40 +391,84 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
-            color=next(colors), plot_control_limits=True, ax=axes[num + 1],
+            color=color, plot_control_limits=plot_control_limits, ax=axes[num],
             palette=None, xtick_interval=xtick_interval)
         c.set_title('{0}: {1}'.format(group_column, group))
+        if spaghetti != 'no':
+            # plot group's sphaghetti on main plot and current subplot
+            for ax in [0, num]:
+                c = _make_spaghetti(
+                    group_md, state_column, metric, individual_id_column,
+                    states, ax=axes[ax], color=color, alpha=0.3,
```
ebolyen: I guess this alpha would become adaptive someday?

q2_longitudinal/_utilities.py
```diff
@@ -400,40 +391,84 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
 
 
+def _make_spaghetti(metadata, state_column, metric, individual_id_column,
+                    states, ax, color=None, alpha=1.0, spaghetti='no'):
+    for ind, ind_data in metadata.groupby(individual_id_column):
+        # optionally plot mean of replicates
+        if spaghetti == 'mean':
+            ind_data = ind_data.groupby(state_column).mean()
+            # add state_column back into dataframe since we use it for plotting
+            ind_data[state_column] = ind_data.index
```
ebolyen: Is there a way to restucture this to avoid mutation? It seems like you could probably pass the data into `_adjust_xticks`, but because this mutates, I'm not sure if there is code relying on this side-effect elsewhere.

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -303,7 +331,80 @@ def test_volatility(self):
...
-            state_column='month', individual_id_column='studyid')
+            state_column='month', individual_id_column='studyid',
+            spaghetti='yes')
+
+    def test_volatility_no_spaghetti_no_control(self):
+        volatility(
+            output_dir=self.temp_dir.name, metadata=self.md_ecam_fp,
+            metric='observed_otus', group_column='delivery',
+            state_column='month', individual_id_column='studyid',
+            spaghetti='no', plot_control_limits=False)
```
ebolyen, line 381: In a perfect world we'd have something about the visualization we could test, but I can't come up with anything, so this is definitely better than nothing.

q2_longitudinal/_utilities.py
```diff
@@ -400,40 +391,84 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
+    cmap = {}
+    num = 1
+    for group, group_md in metadata.groupby(group_column):
+        color = next(colors)
+        cmap[group] = color
         c, gm, gs = _control_chart(
             state_column, metric, group_md, None, ci=ci, legend=False,
-            color=next(colors), plot_control_limits=True, ax=axes[num + 1],
+            color=color, plot_control_limits=plot_control_limits, ax=axes[num],
             palette=None, xtick_interval=xtick_interval)
```
ebolyen: Are the group plots going to have the spaghetti drawn on-top of the control chart? It almost looks like it. Should this section come after the spaghetti so that it matches the draw-order of the global control chart?

q2_longitudinal/_utilities.py
```diff
@@ -400,40 +391,84 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
+                            individual_id_column, ci=95, palette='Set1',
+                            plot_control_limits=True, xtick_interval=None,
+                            yscale='linear', spaghetti='no'):
 
     groups = metadata[group_column].unique()
-    states = metadata[state_column].unique()
-    chart, axes = plt.subplots(len(groups) + 1, figsize=(6, 18))
+    states = sorted(metadata[state_column].unique())
+    fig_count = len(groups) + 1
+    fig_height = fig_count * 6
```
ebolyen: Maybe just inline this into the tuple, then the `6` is less suspicious in context.

q2_longitudinal/_utilities.py
```diff
@@ -400,40 +391,84 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
+        # (e.g., 0, 1, 7, 200 would be plotted at even intervals on x axis)
+        # so that spaghetti aligns with seaborn pointplot x axis.
+        altered_states = _adjust_xticks(ind_data, state_column, states)
+
+        ax.plot(altered_states, ind_data[metric], alpha=alpha, c=color,
+                label='_nolegend_')
+    return ax
+
+
+def _adjust_xticks(ind_data, state_column, states):
```
ebolyen: This name makes me *think* this method mutates something. I don't have a good answer for fixing that, but it's kind of like `_make_*` which definitely mutates.

Would `_get_xticks` work better? I would assume I could call a helper like that multiple times which implies it doesn't mutate anything.

ebolyen (changes requested):
Some minor changes which would probably help readability

q2_longitudinal/_utilities.py
nbokulich: The current order does not seem to impact readability (see below), and I seem to recall running into a bug when I tried switching up the order previously (I believe it had something to do with seaborn messing up the axis spacing).
image

q2_longitudinal/_utilities.py
nbokulich: probably yes — I do not want to introduce too many parameters at this stage, though

q2_longitudinal/_utilities.py
nbokulich: yes! Very happy not to have those clunky default circles.

q2_longitudinal/tests/test_longitudinal.py
nbokulich: Yeah... there might be some way to extract the line coordinates or something but it would probably get really ugly.

q2_longitudinal/_utilities.py
nbokulich: 👍

q2_longitudinal/_utilities.py
ebolyen: oh, they are the same color, so it doesn't matter! sounds good

q2_longitudinal/_utilities.py
nbokulich: I think I'm not following... the mutation is intentional here, to average the data before passing to `_get_xticks` if "mean" is desired. I could do this differently, but with what seems to me like unnecessary extra code (e.g., add an `else` statement)

The mutation can't cause issues elsewhere, since `ind_data` is generated and consumed only within this `for` loop.

But perhaps I misunderstand.

q2_longitudinal/_utilities.py
nbokulich: 👍

q2_longitudinal/_utilities.py
nbokulich: 👍

nbokulich: actually... this appears to not work (the colors are not ordered correctly). I already cycle through the colors in the following `for` loop, so I think it makes sense to just assign the colors in there (making sure the same color order is used) rather than dict comprehension.

nbokulich:
thanks @ebolyen ! I made all changes except for two, as I noted above — pls let me know if more work is needed on those.

q2_longitudinal/_utilities.py
ebolyen: oh sorry, that should be a `collections.OrderedDict` to make the order deterministic. And then ideally you would index that via `group` wherever you previously used `color`. I was hoping we could reduce the number of active "iterables" in the loop.

q2_longitudinal/_utilities.py
ebolyen: I had missed the lifetime of `ind_data` on the first pass, you're right that this probably isn't so bad.

But I think you could simplify things a bit if you passed `ind_data.index` (this is a pd.Index object right?) directly to `_get_xticks` (and `ind_data[state_column]` otherwise) which would also make `_get_xticks` simpler (or even kind of unnecessary) as it doesn't need the entire dataframe for anything, just the series (and `states`).

Here's a potential way to reduce it (I think) and makes `_get_xticks` unecessary:
```python
altered_states = ind_data[state_column]
if spaghetti == 'mean':
    altered_states = ind_data.groupby(state_column).mean().index
if states is not None:
    altered_states = altered_states.apply(states.index)
```
I am pretty sure the above is equivalent to what is here.

q2_longitudinal/_utilities.py
nbokulich: got it, thanks for pointing out the active iterables issue  — `dict` is fine, I just need to call `cmap` from within the loop instead of cycling the colors! :man_facepalming:

q2_longitudinal/_utilities.py
nbokulich: nice... I needed to rearrange and tweak this a bit (your suggestion discards the mean metric), but this definitely simplifies things.

nbokulich:
thanks @ebolyen ! I sorted everything out — let me know if this addresses your suggestions.

q2_longitudinal/_utilities.py
ebolyen: I see what happens now, I missed the reassignment of `ind_data` being super important, which makes sense in retrospect. The reader just needs to understand the API of `df.groupby`'s result  (which I hadn't yet).

ebolyen (approved):

ebolyen:
Thanks @nbokulich!

PR #70, opened 2017-11-02
added static distance/difference computation and tests

### LINKED ISSUE #68: new methods: longitudinal distance/difference from baseline
or other specified time point

ΔYt=Yt−Y0

nbokulich:
fixes #68 

I added these as parameters to `first-differences` and `first-distances` instead of adding new methods, because the methods would otherwise be identical.

Is it too confusing to toggle "static distances/differences" with the `baseline` parameter? @gregcaporaso please also advise (you suggested this functionality, originally as a parameter)

### DIFF
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -714,7 +714,8 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
 
 
 def _first_differences(metadata, state_column, individual_id_column, metric,
-                       replicate_handling='error', distance_matrix=None):
+                       replicate_handling='error', baseline=None,
+                       distance_matrix=None):
 
     # let's force states to be numeric
     _validate_is_numeric_column(metadata, state_column)
@@ -728,14 +729,34 @@ def _first_differences(metadata, state_column, individual_id_column, metric,
     pairs_summary = pd.DataFrame()
     errors = []
     states = sorted(metadata[state_column].unique())
+
+    # if calculating static differences, validate baseline as a valid state
+    if baseline is not None:
+        # validate baseline state
+        if baseline not in states:
+            raise ValueError(
+                'baseline must be a valid state: {0} is not in {1}'.format(
+                    baseline, states))
+        # otherwise splice out baseline state and peg to start of states list
+        # to make iterating over the list easier.
+        else:
+            states.remove(baseline)
+            states.insert(0, baseline)
+
     # iterate over range of sorted states in order to compare sequential states
     for s in range(len(states) - 1):
+        # define whether to calculate first-differences (dY(t) = Y(t) - Y(t-1))
+        # or calculate static differences (dY(t) = Y(t) - Y(baseline))
+        if baseline is not None:
+            state_1 = 0
+        else:
+            state_1 = s
         # get pairs of samples at each sequential state
         group_pairs, error = _get_group_pairs(
             metadata, group_value='null',
             individual_id_column=individual_id_column,
             group_column=group_column, state_column=state_column,
-            state_values=[states[s], states[s + 1]],
+            state_values=[states[state_1], states[s + 1]],
             replicate_handling=replicate_handling)
         # compute distance between pairs
         if distance_matrix is not None:
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -732,6 +732,8 @@ def _first_differences(metadata, state_column, individual_id_column, metric,
 
     # if calculating static differences, validate baseline as a valid state
     if baseline is not None:
+        # cast baseline to same type as states
+        baseline = metadata[state_column].dtype.type(baseline)
         # validate baseline state
         if baseline not in states:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


jairideout:
Would renaming the methods help? e.g. `baseline-differences` / `baseline-distances` or something? I wonder if the current names are too specific.

nbokulich:
thanks @jairideout ! Those were the names I was considering if I split these into separate methods. I think that `baseline-differences` etc would still be too specific (e.g., doesn't capture the functionality of `first-differences`. Something much vaguer like `repeated-differences` could work?

jairideout:
A name that's more vague could work, though it might not be as clear to users what the method does. What about renaming `first-differences` etc to `sequential-differences`, and then adding new methods for `baseline-differences` etc?

I don't have strong feelings here or a good sense of what will make the most sense to users -- interested in what @gregcaporaso thinks about naming and whether to split the method or parameterize it.

q2_longitudinal/_utilities.py
```diff
@@ -728,14 +729,36 @@ def _first_differences(metadata, state_column, individual_id_column, metric,
     pairs_summary = pd.DataFrame()
     errors = []
     states = sorted(metadata[state_column].unique())
+
+    # if calculating static differences, validate baseline as a valid state
+    if baseline is not None:
```
thermokarst, line 734:
```bash
$ qiime longitudinal first-differences \
  --m-metadata-file ecam-sample-metadata.tsv \
  --m-metadata-file shannon.qza \
  --p-state-column month \
  --p-metric shannon \
  --o-first-differences shannon-first-differences.qza \
  --p-individual-id-column studyid \
  --p-replicate-handling random \
  --p-baseline foo \
  --verbose
  File "/home/matthew/.conda/envs/q2dev-2017.11/lib/python3.5/site-packages/q2cli/commands.py", line 218, in __call__
    results = action(**arguments)
  File "<decorator-gen-350>", line 2, in first_differences
  File "/home/matthew/.conda/envs/q2dev-2017.11/lib/python3.5/site-packages/qiime2/sdk/action.py", line 220, in bound_callable
    output_types, provenance)
  File "/home/matthew/.conda/envs/q2dev-2017.11/lib/python3.5/site-packages/qiime2/sdk/action.py", line 355, in _callable_executor_
    output_views = self._callable(**view_args)
  File "/home/matthew/src/qiime2/q2-longitudinal/q2_longitudinal/_longitudinal.py", line 229, in first_differences
    replicate_handling, baseline=baseline, distance_matrix=None)
  File "/home/matthew/src/qiime2/q2-longitudinal/q2_longitudinal/_utilities.py", line 736, in _first_differences
    baseline = metadata[state_column].dtype.type(baseline)
ValueError: invalid literal for int() with base 10: 'foo'

  invalid literal for int() with base 10: 'foo'

See above for debug info.
```

This is related to the discussion below about this parameter's type (`Str` vs some kind of numerical type). If the right type was declared in the method registration, this kind of error would be prevented by the framework.

q2_longitudinal/plugin_setup.py
```diff
@@ -265,9 +266,17 @@
         **miscellaneous_parameter_descriptions,
         **shared_parameter_descriptions,
         'metric': 'Numerical metadata or artifact column to test.',
+        'baseline': (
```
thermokarst, line 269: `If a "baseline" value is provided, sample differences at each state are compared against the baseline state, instead of the previous state.`

I initially interpreted this to mean that `baseline` should refer to a metadata category - I am wondering if this could be updated to make it more clear, although see my comment above about this parameter's type.

q2_longitudinal/plugin_setup.py
```diff
@@ -256,7 +256,8 @@
     inputs={'table': FeatureTable[RelativeFrequency]},
     parameters={**miscellaneous_parameters,
                 **shared_parameters,
-                'metric': Str},
+                'metric': Str,
+                'baseline': Str},
```
thermokarst: Why is this a `Str`? My understanding is that this should be some kind of numerical value here, and is probably contributing to my confusion in the description below.

thermokarst (changes requested):
A few questions @nbokulich, please see inline! Thanks!

q2_longitudinal/plugin_setup.py
nbokulich: Thanks for catching! I suppose I was thinking that we might want to accept non-numeric values (e.g., what if a user's "state" category consisted of the following values: `[1,2,3,4,baseline]`?)

But that thinking is hideously erroneous. If only because we already enforce that `state` is a numeric column (otherwise it might not sort correctly).

So now `baseline` is a `Float`

q2_longitudinal/plugin_setup.py
nbokulich: 👍 

I've added some more detail

nbokulich:
thanks @thermokarst ! I have made `baseline` a `float` and added more detail to the parameter description to emphasize first and foremost that this is a value in `state_column` — let me know if it is still unclear! 🎾

thermokarst (approved):

thermokarst:
@nbokulich, I also updated the type for `baseline` in the `first_differences` registration --- if this looks good to you just let me know, otherwise this PR is good to go! Thanks!

nbokulich:
awesome, thanks for catching! looks good to me — ready to merge when you are.

PR #86, opened 2018-04-18
ENH: Interactive Volatility Control Charts

### LINKED ISSUE #75: spaghetti sample information mouse-over
spaghetti is great, but (in the words of @gregcaporaso ):
>users are going to want to know which subjects the outlier lines (or any lines, for that matter) in these plots are... For example, you might be able to achieve this with mouse-overs that highlight a specific line and give more information about it including the subject id.

### LINKED ISSUE #79: volatility plots: sundry interactive features
- [ ] zoom
- [ ] toggle on/off :spaghetti:
- [ ] toggle on/off control limits
- [ ] toggle on/off mean group trajectories
- [ ] interactive color palette? Could be emperor style custom palette, or more likely just the selection of palettes currently exposed with the `palette` parameter
- [ ] selection of different metrics and metadata grouping values (akin to `alpha-rarefaction`) could be useful, though this will require more structural changes to how these are handled at input

Actually, [many of the parameters](https://docs.qiime2.org/2018.2/plugins/available/longitudinal/volatility/) for this action could be useful as interactive features. E.g., interactively set x-tick intervals, yscale, xscale, but these are less important than those listed above.

### LINKED ISSUE #82: :spaghetti: plots: MPL -> D3
Blocking #79

### LINKED ISSUE #83: `volatility`: make `group-column` optional
CLARIFICATION: without a `group-column` selected, mean lines should still be drawn, but calculated across all samples rather than aggregating by group. (edited 4/23/18)

### LINKED ISSUE #90: `volatility`: show/hide groups/individuals interactively
**Improvement Behavior**
click to show/hide individuals/groups.

**Current Behavior**
currently, can click on the group legend to show single groups, but this only does one at a time. 

**Proposed Behavior**
Would be very useful to, e.g., drop one or more groups to focus on specific groups for comparison.

**Comment**
same with individuals (spaghetti) but no such feature exists. Would be very helpful to hide all spaghetti but one, for example, to compare an individual's trajectory vs. the group mean.

thermokarst:
Fixes #79 (well, all but zooming, will open a new issue for that)
Fixes #75 ~(need to double-check this, I think the mouseover only works for the mean lines ATM)~
Fixes #82 
Fixes #83
Fixes #90 (?)

---

This PR is a bit large, so I broke it up into sequential commits that build on each other, and will hopefully make review a bit more digestible. Whenever it gets merged, lets squash it all down into one commit. Thanks!

---

If this vega business works out, we should move a lot of this boilerplate over to q2templates. Lets open an issue (if necessary) when this PR is merged.

---

cc @ElDeveloper


thermokarst:
Just a quick note on the coveralls failure - that "decrease" in coverage appears to be a result of removing the previously covered code related to `volatility` v1.0 --- there are no newly "uncovered" lines of code introduced in this PR, this is just the result of removing code that is no longer necessary.

thermokarst:
Here is a [sample QZV](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum-qiime2-org.s3-us-west-2.amazonaws.com%2Foriginal%2F2X%2F9%2F99d56334a023314883eeb4b04a6a4d27ad735539.qzv), based on the code in this PR present as of the time of this comment. The viz is based on the [q2-longitudinal tutorial ](https://docs.qiime2.org/2018.2/tutorials/longitudinal/#volatility-analysis) from the docs.

q2_longitudinal/_longitudinal.py
```diff
@@ -162,41 +168,66 @@ def linear_mixed_effects(output_dir: str, metadata: qiime2.Metadata,
...
-               'Global standard deviation'],
-        name='Volatility test parameters')
-
-    raw_data = metadata[[
-        metric, state_column, individual_id_column, group_column]]
-
-    _visualize(output_dir, plot=chart, summary=summary, raw_data=raw_data,
-               plot_name='Control charts')
+def volatility(output_dir: str, metadata: qiime2.Metadata,
+               default_group_column: str, default_metric: str,
```
nbokulich: perhaps the `default_*` columns should be optional?

in general, clearing these out would simplify the interface, but I do like setting these defaults (saves a few clicks!). Making these optional may be a good compromise.

q2_longitudinal/assets/volatility/index.html
```diff
@@ -0,0 +1,155 @@
...
+    // Configuring Vega-generated bound HTML elements is less than ideal, so we
+    // do some manual cleanup here, dynamically. Order matters here.
+    cleanupEl('#group-column', 'Group column', 'select');
+    cleanupEl('#metric-column', 'Metric column', 'select');
+    cleanupEl('#color-scheme', 'Color scheme', 'select');
+    cleanupEl('#toggle-spaghetti', 'Show spaghetti');
+    cleanupEl('#toggle-error-bars', 'Show error bars');
+    cleanupEl('#toggle-global-mean', 'Show global mean');
+    cleanupEl('#toggle-global-control-limits', 'Show global control limits');
+    $('#toggle-global-control-limits').append('<p>+/- 2x and 3x standard deviations</p>');
```
nbokulich: should we clarify that these are 2x and 3x standard deviations from global mean?

q2_longitudinal/plugin_setup.py
```diff
@@ -212,39 +211,36 @@
...
-                'group_column': Str,
-                'ci': Float % Range(0, 100),
-                'plot_control_limits': Bool,
-                'yscale': Str % Choices(["linear", "log", "symlog", "logit"]),
-                'spaghetti': Str % Choices(["yes", "no", "mean"]),
-                'xtick_interval': Int},
-    input_descriptions={'table': (
-        'Feature table to optionally use for paired comparisons.')},
+    inputs={
+        'table': FeatureTable[RelativeFrequency],
```
nbokulich, line 214: This is where interactive metric selection could get tricky... using a feature table could lead to a massive number of possible metrics to choose from. This will not only be messy on the interface, but might cause this to lag/crash for some users? I will test this out to try and break it.

A few ideas:
1. The most basic — but unsatisfying — solution might be to make "metric" a required parameter and not allow metric selection within the viz if only one metric can be input.
2. Same as above, but accept a list of metrics on input. These would be selectable. That would be nice — there will often be a few metrics that a user would like to plot.
3. Same as 2, but make it optional? i.e., allow selection of all numeric columns as metrics if no list is passed. This is nice because one really nice thing about interactive selection is that you don't need to know the name of your metric (some metric names can be non-intuitive, e.g., pielou's E instead of evenness).

nbokulich: CORRECTION it seems that this may be broken — when I input a feature table, none of that data is available in the viz.

q2_longitudinal/_utilities.py
```diff
@@ -404,130 +402,6 @@ def _regplot_subplots_from_dataframe(state_column, metric, metadata,
...
-        # so that spaghetti aligns with seaborn pointplot x axis.
-        if states is not None:
-            altered_states = altered_states.apply(states.index)
-
-        ax.plot(altered_states, ind_data[metric], alpha=alpha, c=color,
-                label='_nolegend_')
-    return ax
-
-
-def _set_xtick_interval(xtick_interval, states):
```
nbokulich, line 489: This is not a feature that we want to strip away (though there must be a better way to do this in vega rather than keeping this function). 

See the broken qzv I've attached to the review to see why... we either need to support explicit interval setting or add sensible auto-scaling in vega.

q2_longitudinal/_vega.py
```diff
@@ -0,0 +1,811 @@
...
+            {
+                'name': 'showGlobalControlLimits',
+                'value': False,
+                'bind': {
+                    'input': 'checkbox',
+                    'element': '#toggle-global-control-limits',
+                },
+            },
+            {
+                'name': 'meanLineThickness',
```
nbokulich, line 130: can we also add a slider bar for spaghetti thickness?

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -362,71 +329,41 @@ def test_linear_mixed_effects_taxa(self):
...
-            spaghetti='mean')
-
-    def test_pointplot_from_dataframe(self):
-        plot_this = pd.DataFrame(
-            {'time': [0, 1, 2], 'metric': [2, 3, 4]}, index=['a', 'b', 'c'])
-        _pointplot_from_dataframe(
-            'time', 'metric', plot_this, None, ci=95, palette='Set1', ax=None,
-            legend=True, color=None, xtick_interval=None)
-
-    def test_volatility_table_data(self):
```
nbokulich, line 409: since table inputs are currently breaking, maybe we should add a test for this? I appreciate that testing visualizations is a PITA so maybe don't test at the action level, just test that feature data is actually being merged and preserved in the metadata, such that it should be accessible in a qzv?

nbokulich: an easy test might be to just use a feature ID as a `default_metric` since the validation comes after merging. Unless if features are being dropped later, e.g., by vega.

nbokulich (changes requested):
BEAUTIFUL.

I am loving vega, and you have done such an awesome job.

There are a few major gaps, though. I'll summarize here but more detail is given in-line:
1. Table inputs do not appear to merge correctly with the metadata, and hence cannot be used as metrics. 
2. x-tick sorting is broken (we discussed this)
3. x-tick intervals should either be explicitly set or sensibly auto-scaled. This was a feature we added to the static viz due to practical needs...

There are a few other minor features that I think we will need on release. These are mentioned in-line.

nbokulich:
Oh I forgot this one — #83  is a feature that I would like to add in when we release this, if possible

In an interactive context, I suppose there should just be a "NA" option so that the means are averaged across all samples instead of splitting by group.

if it is more than a trivial task, we can save that for another day...

q2_longitudinal/plugin_setup.py
thermokarst: Just tested this with the ECAM table you provided in an out-of-band message (after fixing up the broken table handling) - I don't think these column will be too painful for performance. I like the idea of having an optional list of columns to include in the viz, but how about for now we just point people to `feature-table filter-features` if they run up against a performance issue? Would that work for you?

nbokulich: that works for me — I can also document this in the tutorial

ECAM is not massive, but it's a pretty good size (several hundred samples and SVs?) so is probably a reasonable test case — if it's not causing problems, then I think we are okay for now.

thermokarst:
Thanks @nbokulich! I implemented the changes requested above in [this commit](https://github.com/qiime2/q2-longitudinal/pull/86/commits/6ecb4c7d2bd348054c47a88ec9bcefb662acf9d6). Please see the commit message for an itemized list of what was changed. Thanks!

PS - I took a stab at addressing #83 by including the sample ids in the "group" dropdown. Let me know what you think!

q2_longitudinal/_longitudinal.py
```diff
@@ -207,15 +217,10 @@ def volatility(output_dir: str, metadata: qiime2.Metadata,
...
-    numeric = metadata.filter_columns(column_type='numeric')
-
-    # Ensure the default_* columns are members of their respective groups.
-    # This will raise a uniform framework error on our behalf if necessary.
-    categorical.get_column(default_group_column)
-    numeric.get_column(default_metric)
-
-    group_columns = list(categorical.columns.keys())
+    # We want to include the original Sample ID in the viz, so reset the index.
+    data = metadata.to_dataframe().reset_index(drop=False)
```
nbokulich: it looks like this adds the sample name, which does not let us track individual spaghetti (sample name corresponds to individual points).

we want `individual_id_column` — which may well be numeric (as it is with ECAM) and hence may not appear in the group category options by default.

nbokulich (changes requested):
Thanks @thermokarst ! I am loving these new changes! (and learning so much about what `Metadata` can do! 😉 )

My comments at this point are mostly at the interface level (I will send an example qzv offline):
1. It looks like time point 0 is being dropped for some reason — could you investigate?
2. To select individual spaghetti, we need `individual_id_column` (not sample ID) to be forced into the group columns — see comment inline
3. The plot legend cuts off labels — for an example, try setting group column to "id" in the example I'm sending. This is a problem both for number of labels and label length. Number of labels we should fix in this PR if possible, but there might not be a good compromise for label length (we don't really want to wrap long labels so cutting off very long labels might be all we can do. In any case, this is a bug we can raise an issue for now and fix later)
4. `individual_id_column` should be added to the mouse-over metadata shown for 🍝

q2_longitudinal/_longitudinal.py
thermokarst: Ah, thanks for the clarification - I completely misunderstood #83. Should I keep this change, in addition to the `individual_id_column`?

gregcaporaso:
@thermokarst, I was testing these plots this morning and noticed that the mouseovers can be a little deceiving. See these screenshots.

Here's the overall plot I'm looking at - the mean presented in the mouseover seems reasonable:

screenshot 2018-04-23 10 41 29

If I hover near the last time point, but not on the last time point, it looks like the mean is the week 0 mean - this is confusing b/c it's very different than what the line suggests. 

screenshot 2018-04-23 10 42 30

If my hover actually hits that last time point, it shows what I would expect:

screenshot 2018-04-23 10 42 34

Could the mouseover show the mean of the closest (before or after) time point, or of the last time point before the location of the mouse? (either of those is fine by me)

thermokarst:
Thanks @gregcaporaso - were you using Chrome? I noticed the tooltips freak out a bit in Chrome, but not in Firefox. Will take a look - but not sure what I can do - there isn't much we can do to customize this vega plugin. If that is the case, I will need to strip the plugin and implement manually.

q2_longitudinal/_longitudinal.py
nbokulich: My fault, #83 does not have enough description. The goal is to calculate mean lines across all samples, rather than splitting by group. Your change does 50% of that — there is no group, but there is no mean line, either. So we need a way to group all samples together for calculating a "group" mean, perhaps by generating a new metadata column (though that risks overwriting an existing column that happens to be named something like "NA").

For some reason I thought that this change was actually targeting #90 — with the `individual_id_column` change added in you can claim that issue too! or not, if you just want to get this off your plate — that was not in the original manifest for this PR so I would understand.

thermokarst:
Just to clarify - "this vega plugin" == https://github.com/vega/vega-tooltip

thermokarst:
Thanks @nbokulich! https://github.com/qiime2/q2-longitudinal/pull/86/commits/0ae8ab899c4de0f183fed721185a66def65ad862

1. We discussed out-of-band, but for posterity, will record here. This has to do with framework-level metadata merging and missing sample IDs - we won't worry about that here.
2. Replaced the individual samples with the `individual_id_column`
3. After much blood, sweat, and tears, I think we came up with a solution here. The diff is remarkably trivial.
4. Will follow up with another commit or two for addressing @gregcaporaso's mouse-over/tooltip concerns. Stay tuned! 📺

thermokarst:
@gregcaporaso & @nbokulich - I [updated](https://github.com/qiime2/q2-longitudinal/pull/86/commits/9dc88ee0de12fb0aa7dc30f2adb9176d3b2c4f06) the Vega libraries to use the latest and greatest, which includes a handful of improvements w.r.t. tooltips. I also found that by adding in some explicit widgets for controlling the scatterplot symbols (that were previously invisible), it makes it a bit easier to interactively interrogate discrete points.  @gregcaporaso, the reason you were seeing that weird time-point-zero tooltip was because you were first hovering over one of the invisible symbols (good), and then you were hovering over the line, which shouldn't have been displaying a tooltip at all, but was for some reason (bad). This looks like it is fixed now with the latest `vega`.

Check out this [sample viz](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum-qiime2-org.s3-us-west-2.amazonaws.com%2Foriginal%2F2X%2F8%2F8d08fa07c57a22345201febf8eb78f358fcced44.qzv) for an example, and please let me know how it goes. Thanks!

nbokulich (approved):

nbokulich:
Thanks @thermokarst ! These are beautiful!

Everything LGTM. I have failed to break these with my tests. Ready to merge once @gregcaporaso gives his approval.

gregcaporaso:
Looks great @thermokarst, thanks for figuring that issue out! I think that'll save us some support questions on the forum. I'm happy with this version, merge away @nbokulich !

PR #104, opened 2018-06-29
individual ID column is now optional

### LINKED ISSUE #103: ENH: `volatility`: make `individual-id-column` optional
this would allow users to plot volatility charts across any numeric metadata column (e.g., in cross-section experiments), not just repeated sampling over time.

The net effect would just be disabling the 🍝 so that only grouped line plots are shown.

nbokulich:
fixes #103 

@thermokarst this seems to work (I can supply manual testing examples if needed) but I will need your review since I am really just taking to that vega spec like an amateur surgeon. 🔪🍝 

Please let me know if there are additional tests needed — it looks like all the tests you put in previously are mostly "does it run" tests... I assume it would error if vega encountered an invalid spec. These changes do not really require any "input validation" type tests.

### DIFF
--- a/q2_longitudinal/tests/test_longitudinal.py
+++ b/q2_longitudinal/tests/test_longitudinal.py
@@ -355,17 +355,34 @@ class TestLongitudinal(TestPluginBase):
             metric='74923f4bbde849e27fc4eda25d757e2a')
 
     def test_volatility(self):
-        # Just a simple "does it run?" test. Not much worth testing in terms
-        # of the rendered output - vega does all the heavy lifting for us.
+        # Simultaneously "does it run" viz test
+        # plus make sure spaghetti maker works
         volatility(
             output_dir=self.temp_dir.name, metadata=self.md_ecam_fp,
             state_column='month', individual_id_column='studyid')
+        html_path = os.path.join(self.temp_dir.name, 'index.html')
+        with open(html_path, 'r') as f:
+            f = f.read()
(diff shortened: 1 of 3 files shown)


nbokulich:
note: looks like these changes cause line plots to be displayed, but not individual points colored by group.

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -361,6 +361,12 @@ def test_volatility(self):
             output_dir=self.temp_dir.name, metadata=self.md_ecam_fp,
             state_column='month', individual_id_column='studyid')
 
+    def test_volatility_no_individual_id_column(self):
```
thermokarst: This PR gets us in the business of conditionally rendering the vega spec - how about we do a check in this test to make sure that the portions of the spec we expect to be missing are in fact gone now. For example, you could ensure that the substring `spaghettiLine` isn't present (I think - you should double check me on that).

thermokarst (changes requested):
Pulled this critter down and put it through the paces - good work spaghetti-surgeon @nbokulich! I have one minor test request, inline. Thanks!

nbokulich:
thanks @thermokarst ! I have modified 2 tests to confirm that the spaghetti maker is working as intended when individual id column is (not) passed! Let me know if you had something else in mind.

thermokarst (approved):

PR #105, opened 2018-07-07
IMP: Update volatility JS error handling

### LINKED ISSUE #110: BUG: volatility: apostrophes in metadata break viz
**Describe the bug**
Apostrophes in metadata cause `volatility` to run without error; however, the viz is broken and the following error appears on the screen instead of some 🍝 :

`Uncaught SyntaxError: Unexpected identifier`

Other special characters do not cause this problem. I tested:
`*(){},\/"&`

**To Reproduce**
Toss some apostrophes in your metadata; see references below.

**Expected behavior**
A visualization should appear, not an error message in the viz.

**Screenshots**
image

**References**
[forum xref](https://forum.qiime2.org/t/volatility-graph-failing/5258)
[possible approach for validating json](https://stackoverflow.com/questions/5508509/how-do-i-check-if-a-string-is-valid-json-in-python/5508552#5508552) metadata fed to vega spec.

thermokarst:
fixes #110 

Depends on https://github.com/qiime2/q2templates/pull/31

### DIFF
--- a/q2_longitudinal/_vega.py
+++ b/q2_longitudinal/_vega.py
@@ -663,11 +663,6 @@ def _render_volatility_spec(data: pd.DataFrame, individual_id: str, state: str,
                 },
             }])
 
-    # This looks weird, but the idea is to render a unique string into the JSON
-    # spec, that way later on when we render the datatable to JSON, we can
-    # just inject the datatable JSON wholesale.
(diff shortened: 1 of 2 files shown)


nbokulich (approved):
pulled this down, kicked the tires, and LGTM. Thanks @thermokarst !

However, I could not get this to break so am not assessing error handling — really just that it fixes #110 (since I have those test data), and that the code changes look reasonable.

PR #111, opened 2018-07-26
BUG: volatility: convert single quotes for valid vega spec

### LINKED ISSUE #110: BUG: volatility: apostrophes in metadata break viz

nbokulich:
fixes #110 

did not add unit tests because this never raised an error to begin with — just would not display in view.

tried adding json validation (with `json.loads`) to validate the spec... but evidently valid vega specs are ruled invalid by `json.loads`. Any other ideas?

### DIFF
--- a/q2_longitudinal/_vega.py
+++ b/q2_longitudinal/_vega.py
@@ -921,5 +921,9 @@ def _render_volatility_spec(data: pd.DataFrame, individual_id: str, state: str,
     rendered_spec = json.dumps(spec)
     rendered_spec = rendered_spec.replace("'", r"\'")
     rendered_data = data.to_json(orient='records')
+    # convert metadata single quotes to unicode
+    rendered_data = rendered_data.replace("'", r'\u0027')
+    rendered_spec = rendered_spec.replace(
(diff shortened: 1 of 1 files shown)


thermokarst:
Thanks @nbokulich - check out the freshly refactored PR in #105 --- @ebolyen & I hashed out an all new strategy that seems to work pretty darn well for this situation!

nbokulich:
sweet! two birds one stone eh? Shall we close this PR?

Issue #94, opened 2018-05-01
ENH: cross-filter spaghetti viz


nbokulich:
viz for #93 needs new cross-filter vega view.

1. spaghetti, full width of screen on top. Selectable metrics would be a list of features input to `volatility` in the pipeline
2. A barplot of feature importance scores for all features, half-width bottom-left. The metric selected in the side bar (for spaghetti) would cause the associated column in this plot to turn red. Mouse hover will display importance value and feature name.
3. summary statistics — ~LME~, perhaps also descriptive stats — would appear in the bottom-right, based on the metric selected above.

PR #113, opened 2018-08-09
MAINT: Refactoring volatility plot spec renderer

thermokarst:
This sets things up for addressing #94 (this PR could also just be a leading commit on a PR for #94, but this felt like it was too much to digest at once 🍔  🌭 ).

[Sample Viz](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum-qiime2-org.s3-us-west-2.amazonaws.com%2Foriginal%2F2X%2Fc%2Fca0415d65c6bbd6abee243b6128fe68e8865b6a4.qzv)

---

Strategy:

- Move control plot into a subplot (note, there might still be some width issues, I propose fixing those in the followup PR that addresses #94). This will allow us to add cross-filter subplots like its going out of style. 👗 
- Fix control subplot height (previously was dynamic, based on viewport height). This will be important when adding more subplots. 📈 
- Split up vega spec into more modular render templates. 📓 The idea here is that this makes the spec a little more digestible, while also keeping conditional rendering logic sane (conditional JSON rendering was starting to look a little gross, especially when I was prototyping #94).
- Adopt a variable-based identifier control system (let's us use flake8 for simple validation). 🥇


thermokarst:
The current staging env file is broken. Probably not worth fixing just to get travis to run here. Merging this PR will force busywork to build a new env file anyway (hint hint @nbokulich).

nbokulich (approved):
Beautiful. I love the new modular spec templates and variable naming system. This will make it much easier to review/modify in the future.

Pulled this down and took it for a spin... looks good.

One question: should `_vega.py` be removed in this PR or are we hanging on to that for some reason?

thermokarst:
Good catch @nbokulich --- removed that file.

PR #116, opened 2018-08-27
BLD: New recipe dependency

thermokarst:
@nbokulich, is it just q2-sample-classifier that was missing here?


nbokulich:
yes, afaik that's just because we are importing the Importances format from q2-sample-classifier. I don't think any other dependencies were added.

nbokulich (approved):
LGTM thanks @thermokarst !

PR #118, opened 2018-08-28
welcome maturity-index to its new home

nbokulich:
Ported over `maturity-index` from `q2-sample-classifier`.

DO NOT MERGE — blocked by pending PR to port sample-classifier type/format registrations to q2-types.


ebolyen (approved):
It looks like this is actually in a mergable state without the types PR. I'll try it and see if our CI systems clears up.

PR #134, opened 2018-12-22
added anova action

nbokulich:
arguably this could go in q2-diversity instead, but I am placing it here because 1) it relies on statsmodels and some of the same formula parsing and viz code used for LME in longitudinal and 2) this action could be useful for some longitudinal analyses as well as more general use.

### DIFF
--- a/q2_longitudinal/_longitudinal.py
+++ b/q2_longitudinal/_longitudinal.py
@@ -16,6 +16,8 @@ import qiime2
 import q2templates
 import warnings
 import biom
+import statsmodels.api as sm
+from statsmodels.formula.api import ols
 
 from ._utilities import (_get_group_pairs, _extract_distance_distribution,
                          _visualize, _validate_metadata_is_superset,
@@ -26,7 +28,7 @@ from ._utilities import (_get_group_pairs, _extract_distance_distribution,
                          _nmit, _validate_is_numeric_column, _maz_score,
                          _first_differences,
                          _summarize_feature_stats, _convert_nan_to_none,
-                         _parse_formula)
+                         _parse_formula, _visualize_anova)
 from ._vega_specs import render_spec_volatility
 
 
@@ -233,6 +235,46 @@ def linear_mixed_effects(output_dir: str, metadata: qiime2.Metadata,
                model_results=model_results, plot=g, summary=summary,
                raw_data=raw_data,
                plot_name='Regression scatterplots', residuals=res)
+
+
+def anova(output_dir: str,
+          metadata: qiime2.Metadata,
+          formula: str,
+          sstype: int = 2) -> None:
+
+    # Grab metric and covariate names from formula
+    metric, group_columns = _parse_formula(formula)
+    columns = [metric] + list(group_columns)
+
+    # Validate formula (columns are in metadata, etc)
+    for col in columns:
+        metadata.get_column(col)
+    metadata = metadata.to_dataframe()[columns].dropna()
+
+    # Run anova
+    lm = ols(formula, metadata).fit()
+    results = pd.DataFrame(sm.stats.anova_lm(lm, typ=sstype)).fillna('')
+    results.to_csv(os.path.join(output_dir, 'anova.tsv'), sep='\t')
+
+    # Run pairwise t-tests with multiple test correction
+    pairwise_tests = pd.DataFrame()
+    for group in group_columns:
+        # only run on categorical columns — numeric columns raise error
+        if metadata[group].dtype == 'O':
+            ttests = lm.t_test_pairwise(group, method='fdr_bh').result_frame
+            pairwise_tests = pd.concat([pairwise_tests, pd.DataFrame(ttests)])
+
+    # Plot fit vs. residuals
+    metadata['residual'] = lm.resid
+    metadata['fitted_values'] = lm.fittedvalues
+    res = _regplot_subplots_from_dataframe(
+        'fitted_values', 'residual', metadata, group_columns, lowess=False,
+        ci=95, palette='Set1', fit_reg=False)
+
+    # Visualize results
+    _visualize_anova(output_dir, pairwise_tests=pairwise_tests,
+                     model_results=results, residuals=res,
+                     pairwise_test_name='Pairwise t-tests')
 
 
 def _warn_column_name_exists(column_name):
--- a/q2_longitudinal/_longitudinal.py
+++ b/q2_longitudinal/_longitudinal.py
@@ -240,7 +240,7 @@ def linear_mixed_effects(output_dir: str, metadata: qiime2.Metadata,
 def anova(output_dir: str,
           metadata: qiime2.Metadata,
           formula: str,
-          sstype: int = 2) -> None:
+          sstype: str = 'II') -> None:
 
     # Grab metric and covariate names from formula
     metric, group_columns = _parse_formula(formula)
@@ -249,6 +249,8 @@ def anova(output_dir: str,
     # Validate formula (columns are in metadata, etc)
     for col in columns:
         metadata.get_column(col)
+    # store categorical column names for later use
+    cats = metadata.filter_columns(column_type='categorical').columns.keys()
     metadata = metadata.to_dataframe()[columns].dropna()
 
     # Run anova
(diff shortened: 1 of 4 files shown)


q2_longitudinal/plugin_setup.py
```diff
@@ -236,6 +240,29 @@
 )
 
 
+plugin.visualizers.register_function(
+    function=anova,
+    inputs={},
+    parameters={'metadata': Metadata,
+                'formula': Str,
+                'sstype': Int % Range(1, 3)},
```
thermokarst: My understanding is that this parameter maps to `typ` in the [statsmodels API](https://www.statsmodels.org/stable/generated/statsmodels.stats.anova.anova_lm.html#statsmodels.stats.anova.anova_lm):

> The type of Anova test to perform

What is we made this more clear to users by calling this param `test_type`, and made it `Str % Choices(['fixed', 'random', 'mixed'])`, where those map to `{'I', 'II', 'III'}` respectively?

Ref: https://en.wikipedia.org/wiki/Analysis_of_variance#Classes_of_models

q2_longitudinal/_utilities.py
```diff
@@ -518,6 +518,36 @@ def _visualize(output_dir, multiple_group_test=False, pairwise_tests=False,
     })
 
 
+def _visualize_anova(output_dir, pairwise_tests=False, model_results=False,
+                     residuals=False, pairwise_test_name='Pairwise t-tests'):
+    pd.set_option('display.max_colwidth', -1)
+
+    if pairwise_tests is not False:
```
thermokarst, line 525: Can you add a unit test that looks at the results of the rendered HTML when this test is performed?

q2_longitudinal/_utilities.py
```diff
@@ -518,6 +518,36 @@ def _visualize(output_dir, multiple_group_test=False, pairwise_tests=False,
...
+        pairwise_tests.to_csv(os.path.join(output_dir, 'pairwise_tests.tsv'),
+                              sep='\t')
+        pairwise_tests = q2templates.df_to_html(pairwise_tests)
+
+    if model_results is not False:
```
thermokarst: Can you add a unit test that looks at the results of the rendered HTML when this test is performed?

q2_longitudinal/_utilities.py
```diff
@@ -518,6 +518,36 @@ def _visualize(output_dir, multiple_group_test=False, pairwise_tests=False,
...
+        model_results.to_csv(os.path.join(output_dir, 'model_results.tsv'),
+                             sep='\t')
+        model_results = q2templates.df_to_html(model_results)
+
+    if residuals is not False:
```
thermokarst: Can you add a unit test that looks at the presence of these files when this test is performed?

q2_longitudinal/plugin_setup.py
```diff
@@ -130,6 +130,18 @@
...
 
+formula_description = (
+    'Formulae will be in the format "a ~ b + c", where '
+    '"a" is the metric (dependent variable) and "b" and "c" are '
+    'independent covariates. Use "+" to add a variable; "+ a:b" to '
+    'add an interaction between variables a and b; "*" to include '
+    'a variable and all interactions; and "-" to subtract a '
+    'particular term (e.g., an interaction term). See '
+    'https://patsy.readthedocs.io/en/latest/formulas.html for full '
+    'documentation of valid formula operators. On command line, '
```
thermokarst: There are other reasons to enclose it quotes (e.g. `*`) --- maybe this should always recommend wrapping in quotes?

q2_longitudinal/_longitudinal.py
```diff
@@ -235,6 +237,46 @@ def linear_mixed_effects(output_dir: str, metadata: qiime2.Metadata,
...
+    # Run anova
+    lm = ols(formula, metadata).fit()
+    results = pd.DataFrame(sm.stats.anova_lm(lm, typ=sstype)).fillna('')
+    results.to_csv(os.path.join(output_dir, 'anova.tsv'), sep='\t')
+
+    # Run pairwise t-tests with multiple test correction
+    pairwise_tests = pd.DataFrame()
+    for group in group_columns:
+        # only run on categorical columns — numeric columns raise error
+        if metadata[group].dtype == 'O':
```
thermokarst: How about we use the [`qiime2.Metadata` API](https://dev.qiime2.org/latest/api-reference/metadata/#qiime2.metadata.Metadata.filter_columns) to preemptively identify the `Categorical` columns?

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -915,6 +915,30 @@ def test_maturity_index(self):
...
+                         columns=['number', 'letter', 'result'],
+                         index=pd.Index(
+                            ['s1', 's2', 's3', 's4', 's5', 's6', 's7', 's8',
+                             's9', 's10', 's11', 's12'], name='id')))
+
+    def test_execution(self):
+        # does it run?
+        self.anova(self.md, 'result ~ letter+number')
+
+    def test_invalid_formula(self):
```
thermokarst, line 977: FWIW, the formula parser explodes when the tilde is missing --- this might be a good initial check to perform:

```bash
qiime longitudinal anova --m-metadata-file ~/data/moving-pictures/sample-metadata.tsv --p-formula 'DaysSinceExperimentStart' --o-visualization ~/Desktop/foo.qzv --verbose
  File "/Users/matthew/src/qiime2/q2cli/q2cli/commands.py", line 274, in __call__
    results = action(**arguments)
  File "<decorator-gen-425>", line 2, in anova
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.6/site-packages/qiime2/sdk/action.py", line 231, in bound_callable
    output_types, provenance)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.6/site-packages/qiime2/sdk/action.py", line 427, in _callable_executor_
    ret_val = self._callable(output_dir=temp_dir, **view_args)
  File "/Users/matthew/src/qiime2/q2-longitudinal/q2_longitudinal/_longitudinal.py", line 246, in anova
    metric, group_columns = _parse_formula(formula)
  File "/Users/matthew/src/qiime2/q2-longitudinal/q2_longitudinal/_utilities.py", line 854, in _parse_formula
    metric = model_desc.lhs_termlist[0].name()
IndexError: list index out of range

  list index out of range

See above for debug info.
```

thermokarst (changes requested):
Thanks @nbokulich --- a handful of questions, comments, and thoughts, inline. Let me know if you have any questions.

thermokarst:
Also, just noticed the plot title is missing:

screen shot 2019-01-28 at 1 49 45 pm

q2_longitudinal/plugin_setup.py
nbokulich: the model classes you've linked to there are different. See [here](https://rcompanion.org/rcompanion/d_04.html) for a description of the types.

We could do `Str % Choices(['sequential', 'marginal', 'partial'])` but the ANOVA test types appear to conventionally use the I/II/III or 1/2/3 nomenclature, so I am a bit wary of mapping these to terms that are unfamiliar to novice and expert users alike! AFAICT, this nomenclature comes from SAS, where it is standard.

thermokarst: > are different. 

I blame their docs:

> The type of Anova test to perform

It should maybe read like:

> Type of sum of squares to use.

---

How about we go with `Str % Choices(['I', 'II', 'III'])`?

nbokulich: that sounds good to me (blaming statsmodels and using the roman numeral nomenclature)

q2_longitudinal/_utilities.py
nbokulich: I've added a test that loads the TSV (which is downloadable from the html). Good enough?

q2_longitudinal/_utilities.py
nbokulich: never false... have removed if statement (but I now validate the TSV that is saved down there, anyway, just to be clear)

q2_longitudinal/_utilities.py
nbokulich: never false, so have removed if statement

q2_longitudinal/plugin_setup.py
nbokulich: 👍

q2_longitudinal/_longitudinal.py
nbokulich: check out what I've done and see if you like it.

nbokulich:
thanks @thermokarst ! I have updated the tests and addressed your other concerns... see what you think.

thermokarst (approved):

PR #141, opened 2019-06-04
MAINT: removes duplicate "each" from plot-feature-volatility docs

ChrisKeefe:
fixes minor typo


nbokulich (approved):
thanks for catching this @ChrisKeefe !

PR #142, opened 2019-07-08
fixed broken dois

nbokulich:
a couple dois were broken, now they are fixed (full URL to doi.org was given, rather than the actual doi alone)

### DIFF
--- a/q2_longitudinal/citations.bib
+++ b/q2_longitudinal/citations.bib
@@ -6,7 +6,7 @@
   number={6},
   pages={e00219-18},
   year={2018},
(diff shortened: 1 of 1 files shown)


thermokarst:
Have fun with that pandas failure, @nbokulich...

nbokulich:
not a pandas failure, it's statsmodels. Fixed the test data in https://github.com/qiime2/q2-longitudinal/pull/143

ebolyen (approved):

ebolyen:
Thanks!

PR #146, opened 2019-08-24
adding importance filtering to feature_volatility

### LINKED ISSUE #144: ENH: `feature-volatility` should only plot the top N features
**Improvement Description**
`feature-volatility` has the habit of generating QZVs that are too large to open if the model selects a very large number of features. We should expose options to limit the number of features displayed, effectively shrinking the default size of these QZVs.

**Current Behavior**
All important features are displayed.

**Proposed Behavior**
Do something similar to `sample_classifier.heatmap`, selecting by count or by importance threshold (could get even more sophisticated and select the number of quartiles to display):
https://github.com/qiime2/q2-sample-classifier/blob/2b2c0b5c4a41c95986edeb15557bfac8885dbc77/q2_sample_classifier/classify.py#L392-L397

nbokulich:
fixes #144 

This PR:
1. Adds a new utility function for filtering a table based on importance scores, quartiles, or feature counts. This function is used in `plot_feature_volatility` and exposed in that action as well as its parent pipeline, `feature_volatility`. Default is to filter 100 most important features, so this should fix the problems described in #144.
2. Adds new unit tests for utility described in item 1.
3. Adds param `missing_samples` to error or ignore missing samples in `plot_feature_volatility`. (this is not in response to an issue nor is it necessarily a bug — this action is usually run as part of a pipeline where this will not be an issue — just being proactive).

### DIFF
--- a/q2_longitudinal/_longitudinal.py
+++ b/q2_longitudinal/_longitudinal.py
@@ -26,7 +26,7 @@ from ._utilities import (_get_group_pairs, _extract_distance_distribution,
                          _regplot_subplots_from_dataframe, _load_metadata,
                          _validate_input_values, _validate_input_columns,
                          _nmit, _validate_is_numeric_column, _maz_score,
-                         _first_differences,
+                         _first_differences, _importance_filtering,
                          _summarize_feature_stats, _convert_nan_to_none,
                          _parse_formula, _visualize_anova)
 from ._vega_specs import render_spec_volatility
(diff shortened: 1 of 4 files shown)


q2_longitudinal/_longitudinal.py
```diff
@@ -418,7 +418,7 @@ def plot_feature_volatility(output_dir: str,
         _validate_metadata_is_superset(importances, table.T)
 
     # filter table, importances based on importance threshold / feature count
-    if importance_threshold is not None and feature_count is not None:
+    if importance_threshold is not None or feature_count is not None:
```
nbokulich: good call, you are correct that should be `or`

q2_longitudinal/_longitudinal.py
```diff
@@ -418,9 +418,8 @@ def plot_feature_volatility(output_dir: str,
         _validate_metadata_is_superset(importances, table.T)
 
     # filter table, importances based on importance threshold / feature count
-    if importance_threshold is not None or feature_count is not None:
```
thermokarst: @nbokulich, i dropped the guard here, since feature_count can never be `None`, and more specifically, the `_importance_filtering` util is built in a way that can allow it to no-op if necessary, makes sense to just pass on through.

thermokarst (approved):

PR #149, opened 2019-11-13
fix pairwise-difference no variance

### LINKED ISSUE #148: pairwise-differences: t-value referenced before assignment
**Bug Description**
the t-value in pairwise-differences is not being assigned properly when there is no variance between groups. These lines should be amended:
https://github.com/qiime2/q2-longitudinal/blob/8bf27af74a9ff43844ab860132b051098367dcf3/q2_longitudinal/_utilities.py#L241-L244

Probably to assign `t = p = np.nan` instead of `pass`

**References**
See [forum xref](https://forum.qiime2.org/t/longitudinal-pairwise-difference-error/12389/7?u=nicholas_bokulich)

nbokulich:
fixes #148

### DIFF
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -232,6 +232,7 @@ def _compare_pairwise_differences(groups, parametric=False):
     pvals = []
     stat = 'W (wilcoxon signed-rank test)'
     for name, values in groups.items():
+        t = p = np.nan
         try:
             if parametric:
                 t, p = ttest_1samp(values, 0.0)
@@ -240,7 +241,7 @@ def _compare_pairwise_differences(groups, parametric=False):
                 t, p = wilcoxon(values)
         except ValueError:
             # if test fails (e.g., because of zero variance), just skip
-            t = p = np.nan
+            pass
         pvals.append((name, t, p))
     result = pd.DataFrame(pvals, columns=["Group", stat, "P-value"])
     result.set_index(["Group"], inplace=True)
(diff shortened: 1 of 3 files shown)


q2_longitudinal/_utilities.py
```diff
@@ -240,7 +240,7 @@ def _compare_pairwise_differences(groups, parametric=False):
                 t, p = wilcoxon(values)
         except ValueError:
             # if test fails (e.g., because of zero variance), just skip
```
thermokarst: Drive-by Comment:

Don't forget to update or remove this comment.

nbokulich: I left in intentionally... I see assigning `np.nan` as still skipping, just in a more functional way, so I think this comment still applies but happy to update if you have a suggestion.

thermokarst: Oh I think I see now, although that wasn't clear to me at first. How about restructuring a bit? Revert back to the `pass` here (keep the comment as-is), and move the `t = p = np.nan` business up just before the `try` statement. This proposed approach handles the null-case straight off, which I think has a few benefits.

thermokarst (changes requested):
Can we get a unit test to cover the case identified in #148?

q2_longitudinal/tests/test_longitudinal.py
```diff
@@ -328,6 +328,17 @@ def test_pairwise_differences(self):
             individual_id_column='studyid', metric='observed_otus',
             replicate_handling='drop')
 
+    def test_pairwise_differences_zero_variance(self):
```
thermokarst, line 331: This test fails if the fix above is not in place. TDD!

thermokarst (approved):

thermokarst:
@nbokulich, how does that test look to you?

nbokulich:
LGTM thanks!

PR #154, opened 2020-05-27
HACK: Remove "Open in Vega Editor" from volatility for now

### LINKED ISSUE #153: Volatility Visualizations not Appearing Correctly in Vega Editor
**Bug Description**
Volatility visualizations no longer appear correctly in the vega editor likely due to an update (or updates) to vega breaking our templating.

**Steps to reproduce the behavior**
1. Go to https://docs.qiime2.org/2020.2/tutorials/longitudinal/#volatility-analysis
2. Click on 'view' next to 'volatility.qzv'
3. Click on 'Open in Vega Editor'
4. Witness the no graph
5. (optional) resize your browser window and observe the smooshed graph that should appear

**Expected behavior**
The graph renders in the correct proportions, and with no need to resize the browser window

**Screenshots**
Pre-resize no graph
___
Vega Editor - Chromium_037
Post-resize smooshed graph
___
Vega Editor - Chromium_038

**Computation Environment**
- OS: Linux Mint 193.
- QIIME 2: 2020.2
- Web Browser: Chromium, but Firefox experiences nearly identical behavior

**Comments**
1. On Chromium a single resize of my browser window takes the width from 0 to 203 according to the 'width' text box. subsequent resizings of the window change nothing. On Firefox it goes from 0 to 218, and sometimes subsequent resizings do change the width further.
2. I notice the doodads on the bottom to render without resizing they're just all smooshed up. Additionally I can't actually manually enter a width on either browser though the other controls may be interacted with as normal on both browsers.
3. Changing the schema version from 4.0 to 5.12.3 changes nothing about the behavior, but does get rid of the warning.
4. This issue may have something to do with the `autosize` field. At any rate that appears to be what allows the plot to resize so that it shows up after a window resize. I was unable to produce a satisfactory result by messing with this field alone, but it did substantially alter the behavior observed.

**References**
1. https://forum.qiime2.org/t/vega-editor-not-working/15026

Oddant1:
band aids #153 
band aids #155 

1. The plot in vega no longer automatically resizes when the window resizes. I don't know if the vega feature itself is broken now, or if something just broke our previous usage of it, but auto resizing the plot on window resize was just causing the plot to grow wider and wider and wider to unreasonable widths.
2. The plot outside of the vega editor is now too wide. This has to do with the autosize being changed from `fit-x` to `pad`. On `fit-x` it was impossible to effectively resize the plot within the editor. I think if I just reduce the default width this issue may be resolved.
3. This was happening prior to this PR, but the controls on the bottom of the plot in the editor are all smashed together. I'm not sure why. No matter how much width you give the plot they don't seem to separate.

EDIT: In regards to number 2, the plot being too wide or not outside the vega editor is now entirely dependent on monitor resolution. This is obviously not ideal.


Oddant1:
Maybe the answer, at least in the short term, is to leave the autosize on `fit-x` and not have the text box there to change the width. This also allows us to leave the resize on window size change which is definitely a good thing. I'm really not finding any way to keep things sized nicely and consistently while also allowing the user to edit the width.

thermokarst:
> while also allowing the user to edit the width.

We aren't really looking to allow that - the textbox is just a by-product of how we were doing dynamic sizing via signal. Once upon a time that was the best way to handle dynamic resize in vega. I _think_ that some time later in the 4.x or 5 series of vega they added new options to vega for including a signal directly in the autosize attribute. I bet there's just a better, more succinct way to do this, now.

Oddant1:
@thermokarst in that case, my life just got a whole lot easier. Do we still want to leave all those other modifiable settings down there? They don't seem to be breaking anything. And I'm pretty sure I can ditch that textbox pretty easily now that it's not useful.

thermokarst:
I think we need everything listed below, _except_ width:

Screen Shot 2020-05-28 at 11 45 35 AM

Oddant1:
https://github.com/qiime2/q2-longitudinal/pull/154/commits/211098c66dc03949f5d23993f83f55a98f1df273 This should size the plot pretty well overall. I'll look into making it happen without that textbox. Absolute worst case I can call it something like "width (don't change)" I suppose. Or "width (set autosize type to 'pad' before changing)" those get a bit too long though.

EDIT: Making the name that long actually seriously breaks things :joy:

Oddant1:
https://github.com/qiime2/q2-longitudinal/pull/154/commits/f02690b2d55cb11dfa2b25aa36ed277b8be4461d Turns out you can just remove the binding to the textbox and it still works just fine.

thermokarst (changes requested):
Thanks @Oddant1! I took a closer look today, and Vega 5 is going to require a little bit of engineering effort, which we probably don't have much time for before the release. Can you please:

- update this PR to remove the "open in vega editor" button from the template
- file an issue in this repo that puts "upgrade volatility charts to vega 5" on our radar (feel free to start looking into that, too).
- look around the org for any other vega visualizations and do the same two steps on them (remove open in vega editor; open issue about upgrading specs). I think feature-table summarize and composition ancom are two vega vizs, but there might be more. Please let me know if you have any questions!

Oddant1:
@thermokarst I'm not seeing any other viz's that use vega

thermokarst (changes requested):
This LGTM, but please undo the changes to vega-embed.min.js - thanks!

Oddant1:
@thermokarst I believe the js changes were also rolled back here, but do you want me to also role back the other changes I made while trying to make the viz work.

Oddant1:
@thermokarst I think this one also has the js change reverted. Do you want me to revert the other changes I made when we were trying to get the viz to work too?

thermokarst:
> @thermokarst I believe the js changes were also rolled back here

Sorry, I realize where the caching was coming from - I was navigating to the PR via my notifications. The notification I was clicking on was an old when. Sorry for that!

> do you want me to also role back the other changes

Yes please!

thermokarst (approved):

PR #157, opened 2020-06-30
MAINT: updates for scipy 1.5

thermokarst:


nbokulich:
Just for posterity: these changes are to address changes to wilcoxon test approximation in scipy v1.5.0:
https://docs.scipy.org/doc/scipy/reference/release.1.5.0.html

>The wilcoxon function now computes the p-value for Wilcoxon’s signed rank test using the exact distribution for inputs up to length 25. The function has a new mode parameter to specify how the p-value is to be computed. The default is "auto", which uses the exact distribution for inputs up to length 25 and the normal approximation for larger inputs.

nbokulich (approved):
LGTM, thanks @thermokarst ! I will merge once tests pass

thermokarst:
tests probably won't pass, right? the staging env is set to use the old sscipy.

nbokulich:
oooh right. thanks! will just merge as-is. Thanks!

PR #158, opened 2020-07-27
HACK: Pin mpl and mpl base to 322

Oddant1:


ci/recipe/meta.yaml
```diff
@@ -32,6 +32,8 @@ requirements:
     - q2templates {{ release }}.*
     - q2-sample-classifier {{ release }}.*
     - q2-feature-table {{ release }}.*
+    - matplotlib = 3.2.2
+    - matplotlib-base = 3.2.2
```
thermokarst:
```suggestion
    - matplotlib 3.2.2
    - matplotlib-base 3.2.2
```

thermokarst: Do you think we should do an exact pin like this, @nbokulich? Or should we do something like "less than 3.3"?

thermokarst (changes requested):

ci/recipe/meta.yaml
nbokulich: hmm... better do a "less than" just in case, seems less likely to clash with other plugins.

ci/recipe/meta.yaml
```diff
@@ -32,6 +32,8 @@ requirements:
     - q2templates {{ release }}.*
     - q2-sample-classifier {{ release }}.*
     - q2-feature-table {{ release }}.*
+    - matplotlib < 3.3
+    - matplotlib-base < 3.3
```
thermokarst: https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications

> Remember that the version specification cannot contain spaces, as spaces are used to delimit the package, version, and build string in the whole match specification. python >= 2.7 is an invalid match specification. Furthermore, python>=2.7 is matched as any version of a package named python>=2.7.

```suggestion
    - matplotlib <3.3
    - matplotlib-base <3.3
```

thermokarst (changes requested):

thermokarst (approved):

PR #165, opened 2021-08-10
maint: q2-sample-classifier split api updates

thermokarst:
@adamovanja && @nbokulich - the changes in https://github.com/qiime2/q2-sample-classifier/pull/204 need to be propagated here to this plugin - I'm not sure how you want to proceed, we can either discard the new outputs (like I've done here), or pass them through, which will require a bit of churn to make that happen.

Let me know how you want to proceed!

### DIFF
--- a/q2_longitudinal/_longitudinal.py
+++ b/q2_longitudinal/_longitudinal.py
@@ -495,8 +495,9 @@ def maturity_index(ctx,
         table, metadata=metadata, where="{0}='{1}'".format(group_by, control))
 
     md_column = metadata.get_column(state_column)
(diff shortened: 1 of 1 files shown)


thermokarst:
Unrelated - any guesses as to why the `nmit` unit test started failing? https://github.com/qiime2/q2-longitudinal/pull/165/checks?check_run_id=3294991432#step:4:984

nbokulich:
hey thanks @thermokarst !

I think discarding the new outputs is best as a temporary fix... we can revert later on if anyone ever cares to update this action (it has been on my to-do list for a very long time). 

no clue what is going on with nmit, it is unrelated to these other changes. I can take a look next week when I am back in office...

PR #168, opened 2022-01-17
BUG/REF: use Q() for unusual columns in LME

ebolyen:
This helps with columns such as `day-relative-to-fmt`:

Before:
`observed_features ~ day - relative - to - fmt`

After:
`observed_features ~ Q('day-relative-to-fmt')`

Subsumes original logic to escape the formula's response as well.
Tests were already sufficient to test escaping the response side and still pass.


thermokarst (approved):
clean, i dig it.

PR #169, opened 2022-02-07
PERF: Speed up distance computations

ElDeveloper:
Use Pandas vectorized operations to speed up the validation and
computation of first distances and distances to baseline.

In a dataset with ~6K samples runtime went from >1 minute to under a
second.

---

All tests pass except for `test_first_distances_ecam` which fails because of `replicate_handling='drop'`. As far as I can tell when there's a repeated state the repeated state and the next state are dropped. Which seems odd, but wanted to check if this is expected or if I am misunderstanding something. Seems to me that only the repeated state should be dropped.

For example, when I go to `ecam_map_maturity.txt`, I noticed that sample `10249.C001.07SS` is not in `ecam-first-distances` even though the state for that sample isn't repeated. Same is true for `10249.C001.17SS` which should be included since state `10` is repeated but state `11` isn't.  Any help would be great here.

### DIFF
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -692,15 +692,35 @@ def _vectorized_first_distances(
     output = pd.DataFrame(index=metadata.index[:-1], columns=['Distance'])
     output['Distance'] = distances
 
-    # Using the diagonal of the matrix will include some meaningless distances.
-    # For example the distance between the last sample of subject_j and the
-    # first sample of subject_j+1.  We only keep the distances between subjects
-    # with the same identifier.
+    # The next few columns are helpful to filter the results see below
     output['subject_1'] = metadata.loc[
         output.index, individual_id_column].values
     output['subject_2'] = metadata.loc[
         metadata.index[1:], individual_id_column].values
-    output = output[output['subject_1'] == output['subject_2']]['Distance']
+    output['state_1'] = metadata.loc[
+        output.index, state_column].values
+    output['state_2'] = metadata.loc[
+        metadata.index[1:], state_column].values
+
+    states = metadata[state_column].unique()
+    states.sort()
+    states = np.flip(states)
+    pairs = [(states[i], states[i+1]) for i in range(len(states) - 1)]
+
+    def is_keepable_pair(row, reference):
+        return (row['state_1'], row['state_2']) in reference
+
+    # Using the diagonal of the matrix will include some meaningless distances.
+    # For example the distance between the last sample of subject_j and the
+    # first sample of subject_j+1.  We only keep the distances between subjects
+    # with the same identifier.
+    #
+    # Also, not all the distance pairs are allowed
+    output = output[
+        (output['subject_1'] == output['subject_2']) &
+        output.apply(is_keepable_pair, axis=1, reference=pairs)
+    ]['Distance']
+
     output.index.name = '#SampleID'
     output = output.iloc[::-1]
     return output
@@ -760,14 +780,14 @@ def _first_distances_and_distance_to_baseline(metadata, state_column,
     _validate_is_numeric_column(metadata, state_column)
 
     # combine individual and state to vectorize checks
-    combo_name = _generate_column_name(metadata)
-    metadata[combo_name] = (metadata[state_column].astype(float).astype(str) +
-                            metadata[individual_id_column].astype(str))
+    combo_column = _generate_column_name(metadata)
+    metadata[combo_column] = (metadata[state_column].astype(float).astype(str)
+                              + metadata[individual_id_column].astype(str))
     if replicate_handling == 'drop':
-        duplicated = metadata.duplicated(subset=combo_name, keep=False)
+        duplicated = metadata.duplicated(subset=combo_column, keep=False)
     else:
         # this way of finding duplicates is relevant for "random" too
-        duplicated = metadata.duplicated(subset=combo_name)
+        duplicated = metadata.duplicated(subset=combo_column)
         if replicate_handling == 'error':
             if duplicated.any():
                 def summarizer(group, id_column=None):
@@ -784,9 +804,6 @@ def _first_distances_and_distance_to_baseline(metadata, state_column,
 
     # depending on the strategy selected above filter the table
     metadata = metadata[~duplicated].copy()
-
-    metadata[individual_id_column].value_counts()
-    metadata[state_column].value_counts()
 
     # if calculating static differences, validate baseline as a valid state
     if baseline is not None:
@@ -812,6 +829,8 @@ def _first_distances_and_distance_to_baseline(metadata, state_column,
     else:
         output = _vectorized_first_distances(
             distance_matrix, metadata, state_column, individual_id_column)
+
+    output.dropna(inplace=True)
 
     if output.empty:
         raise RuntimeError(
--- a/q2_longitudinal/_utilities.py
+++ b/q2_longitudinal/_utilities.py
@@ -673,6 +673,155 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
     return _dist
 
 
+def _vectorized_first_distances(
+        distance_matrix: pd.DataFrame, metadata: pd.DataFrame,
+        state_column: str, individual_id_column: str) -> pd.Series:
+
+    # sort sample names in a descending order so distances are labeled
+    # according to the sample_i+1 not sample_i
+    metadata.sort_values(by=[individual_id_column, state_column],
+                         ascending=False, inplace=True)
+
+    # when samples are sorted by subject and state the first distances are
+    # found in the second diagonal
+    distance_matrix = distance_matrix.loc[metadata.index, metadata.index]
+    distances = np.diag(distance_matrix.values, 1)
+
+    # ignore the last element because that represents the first sample
+    # identifier
+    output = pd.DataFrame(index=metadata.index[:-1], columns=['Distance'])
+    output['Distance'] = distances
+
+    # Using the diagonal of the matrix will include some meaningless distances.
+    # For example the distance between the last sample of subject_j and the
+    # first sample of subject_j+1.  We only keep the distances between subjects
+    # with the same identifier.
+    output['subject_1'] = metadata.loc[
+        output.index, individual_id_column].values
+    output['subject_2'] = metadata.loc[
+        metadata.index[1:], individual_id_column].values
+    output = output[output['subject_1'] == output['subject_2']]['Distance']
+    output.index.name = '#SampleID'
+    output = output.iloc[::-1]
+    return output
+
+
+def _vectorized_distance_to_baseline(
+        distance_matrix: pd.DataFrame, metadata: pd.DataFrame,
+        state_column: str, individual_id_column: str) -> pd.Series:
+
+    # with the sorted table, we add an extra column with indices so we make
+    # lookups by index not label, which should be faster
+    metadata.sort_values(by=[individual_id_column, state_column],
+                         ascending=True, inplace=True)
+    index_name = _generate_column_name(metadata)
+    metadata[index_name] = np.arange(len(metadata))
+
+    # sort columns based on the metadata, and defer row sorting to the join
+    # operation
+    distance_matrix = distance_matrix[metadata.index]
+    metadata = metadata.join(distance_matrix, how='left')
+
+    # there's four "utility" columns that need to be ignored:
+    # individual, state, combo, and index
+    def column_getter(frame):
+        loc = slice(4 + frame[index_name][0], 4 + frame[index_name][-1] + 1)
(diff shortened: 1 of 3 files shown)


nbokulich:
thanks @ElDeveloper ! this sounds great.

Re: the test failure with `drop`, this is the intended behavior:
https://github.com/qiime2/q2-longitudinal/blob/639ced89de62b3e3c38d898281f9aaef6c59c248/q2_longitudinal/plugin_setup.py#L91-L95

As the `first-distances` are calculated across each interval, missing timepoints (including if a timepoint is dropped) effectively cause _two_ first distances to drop (at time `t` and time `t+1`). So the examples you gave are correct; these are cases where the first distance at time `t+1` cannot be calculated because the preceding timepoint `t` is missing (because all samples were dropped). So the first distances cannot be calculated at either interval `t-1 <--> t` or `t <--> t+1`.

For these reasons, the "drop" option is rather specific and might only be valuable in some situations. "random" is the more suitable option in most real situations, as it will lead to a random sample being selected at those timepoints so that the first distances at both intervals can be calculated. But it sounds like the test is working as originally intended.

ElDeveloper:
Fair enough, thanks for the explanation! I'll fix the code in that case. Would you also be OK if I changed the documentation in `plugin_setup.py`? Seems like "will discard all replicated samples" is ambiguous on the fact that t and t+1 are dropped.

nbokulich:
> Would you also be OK if I changed the documentation in plugin_setup.py? Seems like "will discard all replicated samples" is ambiguous on the fact that t and t+1 are dropped.

documentation updates are very welcome 😄 

but not that line — a bunch of actions in this plugin share the same parameter descriptions. So this action should either get a unique parameter description, or better describe this behavior in the action description (which IMO would be best, since this is the behavior from any missing samples, not specifically `replicate_handling='drop'`)

ElDeveloper:
Cheers thanks so much.

Would you be able to help me diagnose one more case. I've updated my code, but now I am seeing that other samples are currently being excluded from the expected output. For example: `10249.C002.09SS`, `10249.C002.15SS`, `10249.C002.18SS`, `10249.C005.07SS`, `10249.C008.04SS`, `10249.C008.12SS`, `10249.C009.16SS`, `10249.C011.04SS`, `10249.C014.12SS`, `10249.C018.03SS`, `10249.C018.17SS`, `10249.C021.04SS`, `10249.C021.05SS`, `10249.C021.07SS`, `10249.C021.08SS`, etc. When I look at the mapping file for a few of these, I don't see them being associated with any of the repeated sample pairs that are dropped. Any hints as to why these are dropped from the expected output?

I've highlighted the first few in the context of the mapping file here:

## `10249.C002.09SS`, `10249.C002.15SS`, `10249.C002.18SS`

Screen Shot 2022-02-08 at 9 58 36 AM

##  `10249.C005.07SS`

Screen Shot 2022-02-08 at 9 59 42 AM

## `10249.C008.04SS`, `10249.C008.12SS`

Screen Shot 2022-02-08 at 10 04 02 AM

nbokulich:
Hey @ElDeveloper ,
Those first distances are not being calculated because they are missing the immediately preceding timepoint (i.e., it's not being dropped, rather that timepoint is just missing to begin with).

The first distances intervals are meant to be measured across the same time intervals in each subject. So if you have a table like this (rows are samples, cols are months, cells indicate if a sample was collected at that timepoint):
subject-id|0|1|2|3|4|5|10
----------|-|-|-|-|-|-|-|
A|+|+|+|+|+|+|+|
B|+|||+|+|+|+|

A has all first distances calculated, but B only has distances starting at timepoint 4 (i.e., samples were collected at 0 and 3, but the immediately preceding timepoint is missing). On the other hand, one could drop those columns and all distances would be calculated for each sample in a table that looks like this:
subject-id|0|3|4|5|10
----------|-|-|-|-|-|
A|+|+|+|+|+|
B|+|+|+|+|+|

This is the case above, where `10249.C002.09SS` is a sample at month 8, but month 7 is missing. If month 7 were missing in all subjects, the distance interval would be from months 6->8. But that's not the case here, as the interval is based on all samples in the table.

Does that make sense?

Note: the intervals are evenly measuring across samples, but the intervals are not necessarily even across time. One big enhancement here could be to also expose an interval parameter, to allow `auto` (status quo, variable spacing possible), or at explicit intervals (e.g., to ensure that all distances represent 1 month intervals, to set an explicit lag window, or even to skip shorter intervals). I see this as an ENH not needed now, but maybe based on how your code is set up you already have this implemented and it is easier to expose this now instead of totally refactoring.

ElDeveloper:
Thanks so much for the explanation. I didn't realize all this was going on under the hood.  I like the idea of `auto` vs `shared`? or something like that. Let me rework this current version and I'll see where I get. 👍🏽

ElDeveloper:
@nbokulich all tests are passing and should be ready for review now. I've left out the interval selection out for this pass but it is something we could naturally add in a follow-up PR. Looking forward to hearing your thoughts, I know this sort of code can sometimes be hard to read.

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+        output.index, individual_id_column].values
+    output['subject_2'] = metadata.loc[
+        metadata.index[1:], individual_id_column].values
+    output['state_1'] = metadata.loc[
+        output.index, state_column].values
+    output['state_2'] = metadata.loc[
+        metadata.index[1:], state_column].values
+
+    states = metadata[state_column].unique()
+    states.sort()
```
nbokulich, line 706: why sort then flip? why not just sort descending?

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    # with the same identifier.
+    #
+    # Also, not all the distance pairs are allowed
+    output = output[
+        (output['subject_1'] == output['subject_2']) &
+        output.apply(is_keepable_pair, axis=1, reference=pairs)
+    ]['Distance']
+
+    output.index.name = '#SampleID'
+    output = output.iloc[::-1]
```
nbokulich, line 725: can you pls drop in a comment here — why are you flipping the dataframe again?

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    # there's four "utility" columns that need to be ignored:
+    # individual, state, combo, and index
+    def column_getter(frame):
+        loc = slice(4 + frame[index_name][0], 4 + frame[index_name][-1] + 1)
+        # set index to zero because we set the reference to be -Inf
+        return frame.iloc[0, loc].drop(frame.index[0])
+
+    output = metadata.groupby(individual_id_column).apply(column_getter)
+
+    # When the output of groupby is a single series i.e. when there's only one
```
nbokulich, line 754: would this be better caught with a conditional, e.g., by checking the number of unique individuals?

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    output.index.name = '#SampleID'
+    output = output.iloc[::-1]
+    return output
+
+
+def _vectorized_distance_to_baseline(
+        distance_matrix: pd.DataFrame, metadata: pd.DataFrame,
+        state_column: str, individual_id_column: str) -> pd.Series:
+
+    # with the sorted table, we add an extra column with indices so we make
```
nbokulich, line 733: I am not really following this idea... adding, renaming, and handling this extra column seems to take quite a few lines. Wouldn't it just be easier to `reset_index`? instead of indexing on an artificial new column?

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+        (output['subject_1'] == output['subject_2']) &
+        output.apply(is_keepable_pair, axis=1, reference=pairs)
+    ]['Distance']
+
+    output.index.name = '#SampleID'
+    output = output.iloc[::-1]
+    return output
+
+
+def _vectorized_distance_to_baseline(
```
nbokulich, line 729: I like the idea of merging metadata + distance matrix, then using a groupby by individual_id

but some of the logic involved here seems a bit convoluted. It might be easier + more transparent to drop the slicing by index etc. In the groupby operation you could drop the first row (the baseline) and grab the column name matching that index. The result would be a vector of distances to baseline for that individual.

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    metadata[index_name] = np.arange(len(metadata))
+
+    # sort columns based on the metadata, and defer row sorting to the join
+    # operation
+    distance_matrix = distance_matrix[metadata.index]
+    metadata = metadata.join(distance_matrix, how='left')
+
+    # there's four "utility" columns that need to be ignored:
+    # individual, state, combo, and index
+    def column_getter(frame):
```
nbokulich, line 747: this part is a bit difficult to follow — see my comment above.

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    # combine individual and state to vectorize checks
+    combo_column = _generate_column_name(metadata)
+    metadata[combo_column] = (metadata[state_column].astype(float).astype(str)
+                              + metadata[individual_id_column].astype(str))
+    if replicate_handling == 'drop':
+        duplicated = metadata.duplicated(subset=combo_column, keep=False)
+    else:
+        # this way of finding duplicates is relevant for "random" too
+        duplicated = metadata.duplicated(subset=combo_column)
+        if replicate_handling == 'error':
```
nbokulich, line 791: change to:
```
if duplicated.any() and replicate_handling == 'error':
```
???

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+    _validate_is_numeric_column(metadata, state_column)
+
+    # combine individual and state to vectorize checks
+    combo_column = _generate_column_name(metadata)
+    metadata[combo_column] = (metadata[state_column].astype(float).astype(str)
+                              + metadata[individual_id_column].astype(str))
+    if replicate_handling == 'drop':
+        duplicated = metadata.duplicated(subset=combo_column, keep=False)
+    else:
+        # this way of finding duplicates is relevant for "random" too
```
nbokulich, line 789: where is randomization done? as far as I can tell, this rather takes the `first`

taking `first` is not all that bad either 😁 and could be exposed as a separate option. `random` could just randomly shuffle before `duplicated(keep='first')`

q2_longitudinal/_utilities.py
```diff
@@ -673,6 +673,174 @@ def _nmit(table, sample_md, individual_id_column, corr_method="kendall",
...
+                lambda group: baseline in group[state_column].values)
+        if not have_baseline.all():
+            missing = ', '.join([str(i) for i in
+                                 have_baseline[~have_baseline].index])
+            raise ValueError('baseline must be a valid state. The following '
+                             'individuals are missing a baseline'
+                             f' state value of "{baseline}": {missing}')
+
+        # use -np.inf to always make the baseline the first in the list
+        metadata.replace({baseline: -np.inf}, inplace=True)
```
nbokulich, line 825: this will replace all matching values in the entire df, correct? instead of just the `state_column` column?

nbokulich: this also relates to my question about `_vectorized_distance_to_baseline`... instead of doing a pass over the df to `replace` baseline values now, maybe this could just be done inside that function when grabbing the baseline row?

nbokulich (changes requested):
thanks @ElDeveloper ! looks like a vast improvement 😁 

I have not done a test run yet, but just finished reviewing the code. I have a few comments and questions throughout. Pls let me know what you think!

ElDeveloper:
@nbokulich thanks for the detailed review and apologies for the delay getting back to you on these comments. I'll be doing that in the near future. Thanks so much for considering, and for your patience! ⏳ 🚤

lizgehret:
Hey @ElDeveloper, thanks for all of your hard work here! We are currently doing some PR triage and review right now - are you still in the process of addressing @nbokulich's review comments?

ElDeveloper:
Hi @lizgehret thanks for the heads up. I have almost replied to all the comments (although I haven't posted the review yet), but there's a few where I still need to do some additional work. I'll respond ASAP.

PR #174, opened 2022-08-04
bug: `pairwise-distances` fails when samples without group are in metadata but not distance matrix

### LINKED ISSUE #172: `pairwise-distances` fails when samples without state, group, or individual are in metadata but not distance matrix
**Bug Description**
When running `pairwise-differences` and `pairwise-distances`, I notice a discrepancy in how samples that are present in the metadata but not the data (e.g., alpha diversity for `pairwise-differences`, or a distance matrix for `pairwise-distances`) are handled. These "extra samples" in the metadata are ignored by `pairwise-differences`, but can cause a failure with `pairwise-distances` if they don't contain information in the state, group, and individual columns. Ignoring these "extra samples," as `pairwise-differences` does, is more convenient as it allows the user to maintain a single metadata file, even if there are some samples in there that aren't relevant for the `pairwise-*` analyses. As it stands, I have to comment out or remove the "extra samples" from my metadata file to use it with `pairwise-distances`. 

**Steps to reproduce the behavior**
1. Download the three files used in the [`q2-longitudinal` tutorial](https://docs.qiime2.org/2022.2/tutorials/longitudinal/).
2. Edit the metadata file to add a few extra samples. All columns other than sample-id can be left blank for these. For example, the end of my modified metadata file looks like:

```
10249.C057.11SS	y	277	Cesarean	bd	eb	C	9	9.1	Female	57
control-1
control-2
control-3
```
3. Run the following command - it will succeed. 
```
qiime longitudinal pairwise-differences \
  --m-metadata-file ecam-sample-metadata.tsv \
  --m-metadata-file shannon.qza \
  --p-metric shannon \
  --p-group-column delivery \
  --p-state-column month \
  --p-state-1 0 \
  --p-state-2 12 \
  --p-individual-id-column studyid \
  --p-replicate-handling random \
  --o-visualization pairwise-differences-w-controls.qzv
```
4. Run the following command - it will fail with the following error message: 
```
qiime longitudinal pairwise-distances \
  --i-distance-matrix unweighted_unifrac_distance_matrix.qza \
  --m-metadata-file ecam-sample-metadata-w-controls.tsv \
  --p-group-column delivery \
  --p-state-column month \
  --p-state-1 0 \
  --p-state-2 12 \
  --p-individual-id-column studyid \
  --p-replicate-handling random \
  --o-visualization pairwise-distances-w-controls.qzv

  State 0.0 is not represented by any members of nan group in metadata. Consider using a different group_column or state value.

Debug info has been saved to /var/folders/jp/_7s6840d2gdf_gkg1291p7x00000gn/T/qiime2-q2cli-err-e4aw2h6b.log
```

**Expected behavior**
`pairwise-distances` should ignore those extra samples. 

**Computation Environment**
- OS: macOS Big Sur
- QIIME 2 version: 2022.2.1
- q2-lognitudinal version: 2022.2.0

**Full traceback from the above error**
```
  File "/Users/gregcaporaso/miniconda3/envs/qiime2-2022.2/lib/python3.8/site-packages/q2cli/commands.py", line 339, in __call__
    results = action(**arguments)
  File "<decorator-gen-282>", line 2, in pairwise_distances
  File "/Users/gregcaporaso/miniconda3/envs/qiime2-2022.2/lib/python3.8/site-packages/qiime2/sdk/action.py", line 245, in bound_callable
    outputs = self._callable_executor_(scope, callable_args,
  File "/Users/gregcaporaso/miniconda3/envs/qiime2-2022.2/lib/python3.8/site-packages/qiime2/sdk/action.py", line 453, in _callable_executor_
    ret_val = self._callable(output_dir=temp_dir, **view_args)
  File "/Users/gregcaporaso/miniconda3/envs/qiime2-2022.2/lib/python3.8/site-packages/q2_longitudinal/_longitudinal.py", line 101, in pairwise_distances
    _validate_input_values(metadata, None, individual_id_column, group_column,
  File "/Users/gregcaporaso/miniconda3/envs/qiime2-2022.2/lib/python3.8/site-packages/q2_longitudinal/_utilities.py", line 59, in _validate_input_values
    _validate_state_in_dataframe(
  File "/Users/gregcaporaso/miniconda3/envs/qiime2-2022.2/lib/python3.8/site-packages/q2_longitudinal/_utilities.py", line 68, in _validate_state_in_dataframe
    raise ValueError(
ValueError: State 0.0 is not represented by any members of nan group in metadata. Consider using a different group_column or state value.
```

lizgehret:
Addresses #172


gregcaporaso (changes requested):
Thanks so much for the PR @lizgehret! 

I think it makes sense to provide `group_column` when `_load_metadata` is called in `pairwise_differences` as well (line 47). It looks like the filtering is handled elsewhere there (since that action worked for me) but this issue stemmed from inconsistent metadata handling between the two functions, so loading the metadata in the same way seems like it might avoid future issues with that. Just let me know if I'm missing a reason why we shouldn't do that though. 

Also could you add a couple of unit tests: 
 * One for `_load_metadata` that confirms correct behavior when there are and are not `na` values in `group_column`
 * One that confirms that there is no error arising from pairwise_distances when there are `na` values in `group_column`.

gregcaporaso:
@nbokulich, do you want to review this one too, or should we merge when it's ready?

nbokulich:
hey thanks for pinging. I am traveling this week and next so do not think I can review this in time for the next release! So please move ahead without me. Thanks!

lizgehret:
>    I think it makes sense to provide group_column when _load_metadata is called in pairwise_differences as well (line 47). It looks like the filtering is handled elsewhere there (since that action worked for me) but this issue stemmed from inconsistent metadata handling between the two functions, so loading the metadata in the same way seems like it might avoid future issues with that. Just let me know if I'm missing a reason why we shouldn't do that though.

Thanks for bringing that up @gregcaporaso - this actually brings up an issue that I need to investigate a bit further, which is why `pairwise-distances` was failing while `pairwise-differences` wasn't (with the data you provided in the example above). They both call the same validation helper method `_validate_input_values()` which is where `pairwise-distances` fails. They _should_ both exhibit the same behavior - but I'm probably missing something that is the source of this discrepancy. In any case, I'm going to investigate that further and will circle back once that's sorted out!

lizgehret:
Okay circling back on this @gregcaporaso - I examined both methods in detail, and it looks like you were only missing that error message in `pairwise-differences` because you were including multiple files as `--m-metadata-file` arguments, which calls the `merge_metadata` method, taking only the intersection between the two files (thus removing the IDs without any associated data). So this parameter in `_load_metadata()` will be fine to add to both methods!

I'll work on writing up a few unit tests that validate behavior, and then we should be all set here. 🙂

lizgehret:
Alright so I've added a single unit test that tests the `_load_metadata()` helper function in isolation - this covers functionality for both `pairwise_differences` and `pairwise_distances` methods. We should be all set here, but let me know if there's anything else I've missed!

lizgehret:
So final comment/additions to this PR for posterity - @gregcaporaso and I met this morning and discussed adding two additional unit tests that run `pairwise-distances` and `pairwise-differences` with metadata that contains nans in the group column that will fail if `group_column` is ever removed from `_load_metadata()`. Pending passing CI, we should be all set here!

gregcaporaso (approved):
Looks great, thanks so much for taking care of this one @lizgehret! 

Update: we're looking into one last thing before merge here. We overlooked whether this solution will handle the case where there is a sample that _is_ in the distance matrix, but has a nan in group_column. We think that may produce a confusing error message now (e.g., if it says the sample isn't in the metadata, that'll confuse the user who will look at their metadata and see the sample).

Another update: @lizgehret checked on this issue locally, and confirmed that the actions behave as expected (those samples get filtered out since there is no group data associated with them, even though they are present in the distance matrix).

PR #187, opened 2024-06-12
TEST: more robust testing for volatility plot

Oddant1:
This test may be slightly insane.


Oddant1:
This test is probably not really gonna work out.

Oddant1:
@ebolyen suggested maybe forcing the viz to render as an svg during testing so we can just assert that divs exist

lizgehret (approved):
This is a nice approach @Oddant1, this type of testing will really help us with the rest of our visualizations 🥳

